Issue #16, opened 2014-10-21
 EGLIBC 2.15 complains about functions with attribute warn_unused_result


cr1901:
Compiling with a glibc on Ubuntu (version 2.15) results in a number of warnings (not errors) that are probably harmless, but might be annoying to some users :P. I have listed one of these errors below:

```
pass_1.c:4097:10: warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
```

According to [this SO answer](http://stackoverflow.com/questions/7271939/declared-with-attribute-warn-unused-result), this warning is triggered for functions internally declared with the GCC attribute warn_unused_result, when the functions in question are called and subsequently have their return value discarded. Apparently, best practice is to check the return value in some way for these often misused functions :P.

Provided it's not hiding any potential bugs, this of course can be disabled with the flag -Wno-unused-result.

This is more of an FYI than anything. Like it or not, a number of Linux users of this tool probably use GLIBC and the gcc compiler.

vhelin:
Just for the record, I have access to a Linux computer now (Ubuntu 18.04), and I still get lots of warnings like that today. GCC 7.4.0.

/home/vhelin/wla-dx/pass_1.c:5577:3: warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
   fgets(k, 254, stdin);
   ^~~~~~~~~~~~~~~~~~~~

I think it might be silly to add code that would handle the return value as is there really any point in that? Could we fix some problematic situations by doing that? Just hiding the warning makes more sense I guess...

Neui:
The return value of `gets` can be used to determine whenever there were an error while reading. I think this should be handled.

cr1901:
@vhelin I also think it should be handled. The "common" idiom nowadays is to do a lightweight `try`/`catch`:

```
err = fgets()
if err {
    return -1;
}
```

So we could add that code to shut the compiler up as a stopgap (and worry about making the errors better later?).

vhelin:
Anyone wants to fix this? :D

lhsazevedo:
I'll give it a try!

lhsazevedo:
Apparently, we should check if the return is equal to the count parameter.
`fread` returns the number of bytes readed. If it's different from count, the end-of-file was reached or a read error ocurred. [Source](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fread.html)

Should I check this instead of NULL?

[...]

PR #26, opened 2015-01-26
Added CMake build system for convenient native windows builds.

NeonMan:
I wanted to compile WLA under windows without setting up mysys or make so I ported the build system to CMake. While the main target is creating native binaries under windows, cmake can be used to build WLA under linux too.

Both build systems can coexist since the only change to the source tree are the newly created CMakeLists.txt files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,73 @@
+# Main CMakeLists file
+cmake_minimum_required (VERSION 2.8.9)
+
+# --- Project name ---
+project (WLA-DX)
+
+# --- Explicitly enable C
+enable_language (C)
+
+# --- GCC flags 
+if (CMAKE_COMPILER_IS_GNUCC)
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -ansi -pedantic-errors -Wall -Wextra -lm")
+endif (CMAKE_COMPILER_IS_GNUCC)
+
+# --- Tell MSVC to STFU about _CRT_SECURE_NO_WARNINGS
+if (MSVC)
+    add_definitions( -D_CRT_SECURE_NO_WARNINGS )
+endif (MSVC)
+
+#Make the source tree available to includes
+include_directories(${PROJECT_SOURCE_DIR})
+#Generated tables are created at the top build-tree directory
+include_directories(${CMAKE_BINARY_DIR})
+
+# --- Dependencies ---
+add_subdirectory (opcode_table_generator)
+add_subdirectory (wlab)
+add_subdirectory (wlalink)
+
+# --- WLA executables ---
+set(CFILES main.c parse.c include_file.c pass_1.c pass_2.c pass_3.c pass_4.c stack.c listfile.c)
+
+#GB
+add_executable(wla-gb ${CFILES})
+set_target_properties (wla-gb PROPERTIES COMPILE_DEFINITIONS "GB" )
+add_dependencies(wla-gb GB_table)
+
+#WDC65C02
+add_executable(wla-65c02 ${CFILES})
+set_target_properties (wla-65c02 PROPERTIES COMPILE_DEFINITIONS "WDC65C02" )
+add_dependencies(wla-65c02 WDC65C02_table)
+
+#MCS6502
+add_executable(wla-6502 ${CFILES})
+set_target_properties (wla-6502 PROPERTIES COMPILE_DEFINITIONS "MCS6502" )
+add_dependencies(wla-6502 MCS6502_table)
+
+#MCS6510
+add_executable(wla-6510 ${CFILES})
+set_target_properties (wla-6510 PROPERTIES COMPILE_DEFINITIONS "MCS6510" )
+add_dependencies(wla-6510 MCS6510_table)
+
+#W65816
+add_executable(wla-65816 ${CFILES})
+set_target_properties (wla-65816 PROPERTIES COMPILE_DEFINITIONS "W65816" )
+add_dependencies(wla-65816 W65816_table)
+
+#HUC6280
+add_executable(wla-huc6280 ${CFILES})
+set_target_properties (wla-huc6280 PROPERTIES COMPILE_DEFINITIONS "HUC6280" )
+add_dependencies(wla-huc6280 HUC6280_table)
+
+#SPC700
+add_executable(wla-spc700 ${CFILES})
+set_target_properties (wla-spc700 PROPERTIES COMPILE_DEFINITIONS "SPC700" )
+add_dependencies(wla-spc700 SPC700_table)
+
+#Z80
+add_executable(wla-z80 ${CFILES})
+set_target_properties (wla-z80 PROPERTIES COMPILE_DEFINITIONS "Z80" )
+add_dependencies(wla-z80 Z80_table)
+
+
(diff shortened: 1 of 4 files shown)


cr1901:
As long as you have Visual Studio or the Windows SDK installed, both of which still come with nmake you should be able to run win32cl.bat and generate binaries. The Makefiles are nmake-compatible already. The goal is that if you have a make and a hosted ANSI C compiler, WLA should compile correctly. As long as one sticks to a certain subset of make defined [here](http://sai.syr.edu/~chapin/cis657/make.pdf), compilation should work successfully regardless of make type. Perhaps I should've documented this better.

Curiously, I see -D_CRT_SECURE_NO_WARNINGS. I've no need for that last time I compiled binaries using CL.exe, but I was also using a version that predates VS2012.

In any case, before I accept for merge, can you:
* Ensure the binaries go into the "binaries" top level directory
* Make sure you add the -DUNIX define for Unix builds, and -DWIN32 -DMSDOS for Windows builds- CMake appears to [automatically add the first two](http://www.cmake.org/pipermail/cmake/2009-July/030813.html) based on host type (which is okay when host==target, where one is most likely to use CMake), but -DMSDOS is also necessary for Windows.
* Have a method to distinguish Debug from Release builds .

NeonMan:
I guess simply adding -DMSDOS if building on windows is enough, I'll do.

To distinguish debug from release, CMake allows using a prefix or postfix to the executable names. The executables can also be copied to another directory. Is any of those methods acceptable?

CRT warnings are thrown every time the 'unsafe' printf/fprintf/sprintf functions are used. For portable windows/linux software is a MSVC-specific annoyance so I decided to mute them.

cr1901:
>I guess simply adding -DMSDOS if building on windows is enough, I'll do.
Please add, and I'll accept into the tree. It's actually good that the version number was just incremented, so we are free to now rearrange the directory tree as needed.

>To distinguish debug from release, CMake allows using a prefix or postfix to the executable names. >The executables can also be copied to another directory. Is any of those methods acceptable?
CMake appears to have a switch to manually switch between Debug and Release. I'll just recommend users create the CMake directory structure in the binaries folder instead.

>CRT warnings are thrown every time the 'unsafe' printf/fprintf/sprintf functions are used. For portable >windows/linux software is a MSVC-specific annoyance so I decided to mute them.
See Issue #16. Are the warnings just from USE, or are they complaining that the code doesn't check the return value? In either case, I'll accept for now, and revent

vhelin:
I'll let you guys take care of this - I don't know anything about CMake
myself. :)

On Fri, Jan 30, 2015 at 11:56 AM, William D. Jones <notifications@github.com

NeonMan:
I've been out for a few days. I'll fix CMake config files soon and add a pull req.

NicklausW:
That's most likely because the program uses fopen() and other file functions. I really don't see why it's necessary to use Microsoft's "safe" version fopen_s(), though.

cr1901:
I already accepted the pull request. I'll leave the warning-disable in for now, and figure out whether -DMSDOS is in fact still needed soon.

cr1901:
NicklausW: win32.bat is meant for CL.EXE, either provided by the Windows SDK or Visual Studio. unix.sh might've worked okay, but with MinGW the intent was to run win32, which is in fact a shell script.

In any case, I improved the CMake support. I think we can safely remove the win32, unix.sh, and win32cl.bat scripts, as well as the corresponding Makefiles. For the "canonical" Windows and POSIX (including those not limited to Apple, BSD, or Linux- maybe AIX users, for instance :P) users, just require CMake.

I feel we should keep the remaining Makefiles for those who need non-traditional platforms (I still may remove the templates from the main branch and/or move them to another branch). Any objections, Ville?

vhelin:
On Mon, Feb 9, 2015 at 3:17 PM, William D. Jones <notifications@github.com>
wrote:

> NicklausW: win32.bat is meant for CL.EXE, either provided by the Windows
> SDK or Visual Studio. unix.sh might've worked okay, but with MinGW the
> intent was to run win32, which is in fact a shell script.
>
> In any case, I improved the CMake support. I think we can safely remove
> the win32, unix.sh, and win32cl.bat scripts, as well as the corresponding
> Makefiles. For the "canonical" Windows and POSIX (including those not
> limited to Apple, BSD, or Linux- maybe AIX users, for instance :P) users,
> just require CMake.
>
> I feel we should keep the remaining Makefiles for those who need
> non-traditional platforms (I still may remove the templates from the main
> branch and/or move them to another branch). Any objections, Ville?
>

No objections. As long as all works. :)

cr1901:
@vhelin Do you still have an Amiga 3.0/4.0 machine or can bring up a VM to test? The legacy branch has been stagnant for a while, but I would like the old binaries to still work/test them. I'd like to bring the old targets up to date and delete the legacy branch.

A number of utils for game dev (Deluxe Paint comes to mind) still only exist on Amiga, OS/2, DOS etc machines. Chances are, alternatives aren't going to really pop up (ask me how I feel about pixel art programs sometime). So still feel it's important that these targets are supported so development is done in one place. 

Cmake is *not* going to work for a native compile to these systems- it is not possible to generate a Cmake makefile that doesn't depend on binaries that Cmake provides. I'm unaware of cross compilers for the systems. However, because of CMake, it is no longer important that Makefiles be portable- as long as the code is.

For legacy systems, we could create a "legacy" top level directory, and a directory for each legacy non-Cmake system which contains a usable makefile/build script for that directory *as well as the ability to run tests* (we can discuss this later). The main tree is left alone.

WLA's always been pretty fast. Now that hash tables are used, I'm guessing the compiling time will still be negligible on slow machines.

vhelin:
On Tue, May 31, 2016 at 7:05 PM, William D. Jones <notifications@github.com>
wrote:

> @vhelin <https://github.com/vhelin> Do you still have an Amiga 3.0/4.0
> machine or can bring up a VM to test? The legacy branch has been stagnant
> for a while, but I would like the old binaries to still work/test them. I'd
> like to bring the old targets up to date and delete the legacy branch.
>

I don't have WinUAE installed on my new computer, but could do that after I
return home in 10 days. Don't have real hardware, sold my A1200 like 10
years ago away to a friend... :)



> A number of utils for game dev (Deluxe Paint comes to mind) still only
> exist on Amiga, OS/2, DOS etc machines. Chances are, alternatives aren't
> going to really pop up (ask me how I feel about pixel art programs
> sometime). So still feel it's important that these targets are supported so
> development is done in one place.
>
> Cmake is *not* going to work for a native compile to these systems- it is
> not possible to generate a Cmake makefile that doesn't depend on binaries
> that Cmake provides. I'm unaware of cross compilers for the systems.
> However, because of CMake, it is no longer important that Makefiles be
> portable- as long as the code is.
>
> For legacy systems, we could create a "legacy" top level directory, and a
> directory for each legacy non-Cmake system which contains a usable
> makefile/build script for that directory *as well as the ability to run
> tests* (we can discuss this later). The main tree is left alone.
>

Sounds good! And yeah, it would be interesting to see the current state of
e.g. AmigaOS builds...


> WLA's always been pretty fast. Now that hash tables are used, I'm guessing
> the compiling time will still be negligible on slow machines.
>

I remember WLA DX being slower than what I wanted, on my old Amiga 1200,
68030/50, but then again back in the day, when I did the comparison, I was
still asm coding... :) And yeah, the hash tables should speed up things a
lot! But lets see. If nobody else tests this, I'll give it a go with WinUAE
when I get back home next month.

PR #35, opened 2015-06-29
Partially fixed error with 6502/65c02/6510/65816/spc700 opcodes, and improved win32 builds.

nicklausw:
To be honest I'm not sure if this is even worth a pull request (it's two separate fixes to merge in,) but just felt like my commits should be apart of the main thing.

The first fix was making it so that all 6502/blablabla opcodes don't complain when there isn't an operand hint. That's only a partial fix; labels still have the problem, which is just handled by the '.16bit' directive.

The second fix was just adding a batch file for Windows that actually properly outputs binaries without Visual Studio or whatever. It helps me a whole lot (as I'm more comphy with Windows than Ubuntu,) and will probably help others who need binaries for a sudden fix and aren't on some form of Unix.


cr1901:
Yea, I made a mess of the build system, but... the old Makefiles are currently only relevant for MSDOS and Amiga. For any platform that runs CMake, they are completely out of data/I should be completely remove them if it weren't for compatibility, tbh.* So I'll probably have to cherry pick when commits I want.

I won't have time to look over your changes for a while, so I'll leave the pull request open for now. I can envision times where one wants a 16-bit label when WLA will try to use an 8-bit label if within range (my personal use case is trying to make a disassembly for eg. SNES games, which can be reassembled into a bit-compatible image). So the hints and .16BIT directive may still be useful there.

*That, and adding a custom CMake target is not enjoyable.

nicklausw:
...Can I ask that one of you at least cherry pick the commits involving the operand hinting mess?

cr1901:
I don't actually know how to cherry-pick, so I've been putting it off XD.

I'm not in a position to correctly judge whether your fix is correct or not- I actually don't understand the code base nearly as well as Ville, and have commit access to tune up the code base, among other reasons. I defer to Ville's judgment on this.

Re: the Windows Makefiles. The Windows target now uses CMake- the old Makefile mess that I have for Win32 really needs to be deleted, and the old Makefiles are kept for Amiga/MSDOS compat (I don't believe that they should be left out just because the build system changed). So I really don't do a great job even with the features I add LOL.

nicklausw:
I'm perfectly fine with my little make system being untouched. It's only there because I was too lazy to get Cmake (which works fine).

Don't see what's so hard to understand about the fix. It just automatically gives operand hinting where the assembler oddly forces it. Perhaps the fix could be avoided with some corrections in the operand tables?

cr1901:
It's not that I don't understand the fix... I'm wondering what effects it has on size autodetection (which also needs work) and what Ville's intent was with the code that's currently there.

I'm accepting for now, and will make changes as necessary.

nicklausw:
K then, I suppose that's a good way to go about it.

PR #39, opened 2015-07-23
Minor logic fix in pass_1.c

martinlindhe:


martinlindhe:
Im not 100% this is the correct fix, just getting familiar with the code base..

vhelin:
Good fix! :)

PR #40, opened 2015-07-24
pass_1.c: check for negative value, avoids a crash on some odd binary…

martinlindhe:
… sequences

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,8 +27,10 @@ if (UNIX)
     #http://www.cmake.org/pipermail/cmake/2009-April/028439.html
 endif (UNIX)
 
-SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ansi -pedantic-errors -Wall -Wextra")
-string(REPLACE "-O2" "-O3" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
+if ((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID MATCHES "GNU"))
+    SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ansi -pedantic-errors -Wall -Wextra")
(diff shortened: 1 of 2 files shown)


martinlindhe:
patch updated. after fixing cmake, yes clang now catches the error:

```
/Users/m/dev/wla-dx/pass_1.c:741:9: error: ISO C90 forbids mixing declarations and code
      [-Werror,-Wdeclaration-after-statement]
    int op_id = tmp[0];
        ^
1 error generated.
```

thanks for the pointers!

cr1901:
We can't set the flags unconditionally b/c Windows CL.EXE and Watcom won't understand those flags. CMAKE seems to provide a CMAKE_C_COMPILER_ID variable: http://www.cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html#variable:CMAKE_%3CLANG%3E_COMPILER_ID

Can you try this?
```
IF ((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID MATCHES "GNU"))
SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ansi -pedantic-errors -Wall -Wextra")
    string(REPLACE "-O2" "-O3" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
ENDIF ((CMAKE_C_COMPILER_ID MATCHES "Clang") OR (CMAKE_C_COMPILER_ID MATCHES "GNU"))

martinlindhe:
okay sorry. didnt mean to make matters worse :-)

So what targets does are supposed to work with the cmake build? i dont know if it supports watcom compiler, isnt that quite old?

cr1901:
Ideally, any hosted ANSI C compiler with an 8-bit char should be able to compile this; if CMake doesn't support the target, I'll add the relevant local files (or alternatively, provide custom Makefiles).

Watcom is supported in CMake, I just used it as an example.

In any case, just try the code snippet I presented in my previous reply.

martinlindhe:
Okay, i tested it working (detected gcc on Ubuntu 15.04, system clang on OSX 10.10)

cr1901:
Accepting and merging. Thanks for the patch/cooperation!

PR #45, opened 2015-08-03
One big makefile overhaul.

nicklausw:
I'm not sure this is a good idea, but decided to remove the assembler's need to know the OS and pretty much went all-out in cleaning up the makefiles.


nicklausw:
What was I on last night...

cr1901:
You weren't TOO far off what I intended to do with the Makefiles, but I'd prefer to wait a bit.

nicklausw:
The one thing that I did (unfortunately in a huge commit) that I'd still like to have in the source: changing win32_good.bat and makefile.win32.all so that the batch file doesn't have to manually delete all the .o files.

PR #66, opened 2015-11-14
Fixed makefile dependency generation outputting targets after depende…

Denwey:
Fixes dependencies getting outputted like this:
```Makefile
 source.s \ 
    gb.inc
target.o:
```

To this:
```Makefile
target.o: source.s \
    gb.inc
```

### DIFF
--- a/include_file.c
+++ b/include_file.c
@@ -484,27 +484,27 @@ int print_file_names(void) {
   /* handle the main file name differently */
   if (fni != NULL) {
     if (fni->next != NULL || ifd != NULL)
-      fprintf(stderr, "%s \\\n", fni->name);
+      fprintf(stdout, "%s \\\n", fni->name);
     else
-      fprintf(stderr, "%s\n", fni->name);
+      fprintf(stdout, "%s\n", fni->name);
     fni = fni->next;
   }
 
   /* included files */
   while (fni != NULL) {
     if (fni->next != NULL || ifd != NULL)
-      fprintf(stderr, "\t%s \\\n", fni->name);
+      fprintf(stdout, "\t%s \\\n", fni->name);
     else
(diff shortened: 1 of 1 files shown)


cr1901:
I don't use Makefile dependency generation, but my gut feeling is that it's canonical to output dependencies to stderr instead of stdout.

Additionally, stderr and stdout have different guarantees about when they flush, so I'm not currently sure if this is the correct solution to the problem (i.e. it may work on your platform, but may break on others). I won't have time to check until at least next week, but by all means, if it works for your platform/libc that's great :D!

Denwey:
I'm running it under a Debian environment, where more standardized tools like gcc output both targets and dependencies into `stdout`. The patch seems to make wla talk better with make, without it I need to write an explicit recipe because otherwise make will complain with syntax errors, because of the different flush times `stdout` and `stderr` may have.

Moreover, `stderr` should actually be used only for exceptional info the user should know and take action upon. Expected output (makefile dependencies) should get outputted to `stdout`, otherwise the dependency files could occasionally get filled with, say, assembly phase errors which are of no use to make.

Denwey:
I'm running it under a Debian environment, where more standardized tools like gcc output both targets and dependencies into `stdout`. The patch seems to make wla talk better with make, without it I need to write an explicit recipe because otherwise make will complain with syntax errors, because of the different flush times `stdout` and `stderr` may have.

Moreover, stderr should actually be used only for exceptional info the user should know and take action upon. Expected output (makefile dependencies) should get outputted to `stdout`, otherwise the dependency files could occasionally get filled with assembly phase errors which are of no use to make.

vhelin:
Good fix, seems like I was accidentally mixing stderr and stdout here. :P

Issue #60, opened 2015-09-06
Add ability to use environment-variables for file-includes


bazzinotti:
'nuff said... 
For now I just write my projects in a way that I can get around this desire.. ie from a central Makefile and every file is referenced from the directory the makefile is in.

But making projects with many object files and a central Makefile is sometimes a hard habit for me to get in when starting a new project -- it's during these times I wish I could use an environment variable to escape directory dependency...

Anyways -- this flip-flop attitude I have makes me wonder if this feature should even be added (worth the effort to add?) Let's talk about it ---> For one, environment variables are used differently whether on Windows and *nix -- but let's say WLA DX suddenly had the ability to handle environment variables -- do we have a way as App Developer to check the host OS so that we can conditionally include files in a platform independent manner? 

Food for thought...

Extra thought -- What platform independent manner can we include files to begin with using WLA DX (I remember learning this but I've since forgotten :X)

cr1901:
WLA requires a hosted ANSI C compiler, and any conforming C compiler will provide a getenv() that will either retrieve environment variables or return a NULL ptr IIRC. So platforms that support getenv() meaningfully get support.

> Extra thought -- What platform independent manner can we include files to begin with using WLA DX > (I remember learning this but I've since forgotten :X)
You can't. C makes no guarantees re: the format of paths that fopen and friends accept. I don't personally care for the .INCDIR directive and would much rather see them moved to the command line as parameters, similar to POSIX (and NOT an environment variable either). But that would require rewriting the command line parsing. And if we want to get really pedantic, hosted C compilers are not required to meaningfully support argc/argv[] :P.

Let's worry about the bugs first before we start adding new features.

vhelin:
pass_1.c hosts a function as follows:

int localize_path(char *path) {

  int i;

  if (path == NULL)
    return FAILED;

  for (i = 0; path[i] != 0; i++) {
#if defined(AMIGA) || defined(UNIX)
    /* '\' -> '/' */
    if (path[i] == '\\')
      path[i] = '/';
#else
    /* '/' -> '\' */
    if (path[i] == '/')
      path[i] = '\\';
#endif
  }

  return SUCCEEDED;
}

... and that is used with all the file include/incbin functions. So you can write the paths using ///'s or \\\'s, WLA will convert the path to the working format. Is this what you wanted?

bazzinotti:
[...]

PR #68, opened 2015-11-22
Added -I and -L to command line, where one can specify a global include/library directory.

galaxyhaxz:
Should fix github issue #60 

EDIT: I wanted to make it like ANSI-C where root files are specified with 'file.i' or \<file.i\> but this would require working over the parser in WLALINK to require strings. I think how it is currently shouldn't be a problem in most cases however.


vhelin:
Thanks mr. galaxyhaxz! I'll try to check this one out in the next couple of days!

vhelin:
Thanks, looks good! :+1:

PR #72, opened 2015-12-31
Make wla-dx folder

nicklausw:
Cmake builds just fine with this, but I'm unsure about the old build scripts. The legacy branch exists for a reason, I suppose.

Also the merge can't be automatic apparently, dunno about that.


cr1901:
Sorry, for backwards compatibility reasons I'm not going to merge this. Ville is free to override me, however :).

nicklausw:
Thought so. The current layout just bothers me, with there being no src directory for the main program or anything.

What if I modified all the legacy scripts?

vhelin:
Yeah, I also value the backwards compatibility a lot. :) But if you can make it so that all the legacy scripts works, then moving all WLA files to its own folder would be ok.

PR #73, opened 2015-12-31
Implemented GBHEADER

nicklausw:
There's probably mistakes in the code, but all my tests go through just fine.


nicklausw:
Yep, just caught one mistake, GBHEADER doesn't tell complement check to compute. Fixed.

vhelin:
Looks good - thanks! :+1:

Issue #52, opened 2015-09-01
!! #(imm. value) not being parsed correctly in call to macro


bazzinotti:
From my last issue filed today (associated with macros), https://github.com/vhelin/wla-dx/issues/50
Ville wrote: `The assembler itself should find that # and work
accordingly`

This is not the case (huc6280) with such an example: 

```
.macro sub          ; subtract byte-sized value
     sec
     sbc    \1
.endm

; later in a code section
sub   #$20
```

The error is `STACK_CALCULATE: Syntax error. Invalid use of modulo`

The reason this bothers me is because it's blocking WLA DX from being able to use one macro to potentially handle different kind of arguments all in one macro (convenience) (RAM/ROM access instead of immediate value). 
I am hoping to _not_ be recommended a workaround (2 macros). I'd like to see or help WLA DX become awesome -- instead of working around its problems.

bazzinotti:
This is probably the most aggravating bug of them all.. I'm trying to simplify my work with macros but it's irritating to have to create separate macros based on immediate/address knowing full well that if WLA DX was fixed I would never have had to create these separate macros and separate calls to them. 

But I don't know where to start in fixing this issue :[ if anyone doesn't have the time but has the know-how -- a pointer or three or four would be helpful. Thank you

bazzinotti:
This is probably the most aggravating bug of them all.. I'm trying to simplify my work by using macros but it's irritating to have to create separate macros based on immediate value/address access knowing full well that if WLA DX was fixed I would never have had to create these separate macros and separate calls to them. (the .w bug https://github.com/vhelin/wla-dx/issues/53 is also very related -- as it means having to force create different kinds of macros based on their ZP/ABS address access, when this could be automated!)

But I don't know where to start in fixing this issue :[ if anyone doesn't have the time to fix the whole issue but has the know-how -- a pointer or three or four would be helpful. Thank you. Otherwise I will try to wait.

I really want to fix these issues so I can code my engine the right way!! I'm X_X otherwise.

bazzinotti:
bug_exhibition : http://botb.club/~bazz/wladx/issue52.zip

vhelin:
Oh, now I can see why the assembler doesn't find that #... :) It's part of a string argument to a macro! I bet

...

.macro sub          ; subtract byte-sized value
     sec
     sbc    #\1
.endm

; later in a code section
sub   $20

...

[...]

PR #74, opened 2016-01-07
Fixed some invalid modulo cases?

nicklausw:
Should handle #52. The stack calculator says that the case of "#symbol" shouldn't work, but apparently "symbol * #symbol" should work. This fix reverses that.


nicklausw:
Anyone have any idea what's up with the conflicts?

PR #77, opened 2016-01-08
Added ASCII commands to strings and .DL support for 65816

galaxyhaxz:
(COPIED FROM FORK)

ASCII commands work with .DB and .ASC strings. They are as follows:
'\0' -> insert null byte
'\x' -> insert hex character
'\\>' -> set highest bit (0x80) of preceding character
'\\<' -> set highest bit (0x80) of proceeding character
'\\\' -> escape character

.ASC hex characters are NOT remapped. This is useful if you need to write special characters. Example:
.ASC "My special character: "
.DB $59
...becomes
.ASC "My special character: \x59"

'\\\' is an escape character. So if you actually need to write .DB "Testing \x 123" you would instead need to write .DB "Testing \\\x 123" just like in ANSI-C. Thus '\\\\<' and such skips the special command.

'\\<' and '\\>' were added in because in many SNES games, i have seen strings terminated by setting the highest bit. This saves space instead of inserting a zero at every string.

.DL for 65816 works just like you'd expect it. It writes the bank byte when used with labels.

One last note, the .DL support is not added to structs and such yet. I figured I'd get around to this soon enough for full .DL/.LONG support. Keep in mind i'm not going to add .DD/.DWORD because none of the architectures WLA supports use 32-bit addressing. Thus I see no use in .DD support.


galaxyhaxz:
TODO:
- Update readme accordingly
- .DL support for .STRUCT and various other commands
- Octal special character support (000-777) for '\x' command (e.g. .DB "Octal \x472 Hex \xF9")

vhelin:
Thanks! :) Sorry for letting this stay unmerged for so long, I somehow missed & forgot all about it...

Anyway, could you also update the README as well? :+1: 

Question: ASCII commands are now supported in strings given to .DB, but should ASCII commands be supported everywhere?

Issue #81, opened 2016-01-20
Give sections "identifiers" to access labels by


Drenn1:
Okay that's a crummy summary. Basically, to reduce confusion over which labels are in which bank (disassemblies are huge), my thought is to have a way to make a section's labels accessible only by typing "[section identifier].[label name]", or something along those lines. Of course this would only apply outside the section.

```
.SECTION "Bank 15" IDENTIFIER "bank15"

someFunction1:
    ; some code

someFunction2:
    call someFunction1 ; Works

.ENDS

...

; Later, possibly in a different bank
someFunction3:
    call someFunction1 ; Compile error
    call bank15.someFunction1 ; Works
```

PR #82, opened 2016-01-22
Implement "namespaces" for sections

Drenn1:
This is an implementation for issue #81. Though I changed the name to "namespaces" since that's basically what it is. This isn't necessarily complete but I'm hoping to discuss it a bit. It works like this:

```
.SECTION "Bank 15" NAMESPACE "bank15"

someFunction1:
    ; some code

someFunction2:
    call someFunction1 ; Works

.ENDS

...

; Later
someFunction3:
    call someFunction1 ; Compile error
    call bank15.someFunction1 ; Works
```

I included an example for debugging purposes.

Some notes / things I'd like to discuss:
- A .using directive could be appropriate.
- I haven't tested libraries yet.
- The linker will probably get confused if you define labels like "identifier.name" when a label "name" already exists in a section with namespace "identifier". Or something like that.
- Currently .define's inside sections don't have this behaviour, they go into the "global" namespace. I think they should probably be treated the same as labels instead, I just wanted to see what people think first.
- Should ramsections, enums, etc have namespaces? Would it be better to have a separate ".NAMESPACE" directive?
- I see why you were talking about refactoring now!


nicklausw:
So let's say this happens:
```
.section "blabla"
thingy:
xor a
.ends
ld hl,thingy
```

since no namespace was defined, will an error be given?

Drenn1:
No, that's fine, stuff goes into the global namespace if no namespace is specified. My goal isn't to break everyone's existing code. :)

vhelin:
Nice work - looks great! :D I'll take a closer look at this tomorrow when I
have more time...

Drenn1:
I've added a commit which just fixes a typo and removes some irrelevent comments in the example.

Drenn1:
It's looking like my hashmap implementation will need to change this a fair bit. Maybe you should wait a bit and I'll make a pull request for both?

nicklausw:
Now that you mention the lack of error with pre-existing code, I'm all for namespaces.

vhelin:
Ok, no hurry! Take all the time you need to make it good enough. :)

vhelin:
And guys, feel free to add yourselves to section 19 of README ("Thanks
to"). :)

PR #88, opened 2016-02-14
Added .NINTENDOLOGO for gameboy roms

nicklausw:
Rather than have to manually put in the Nintendo logo for gameboy roms written for wla, you can now just do `.nintendologo` instead.

While this should be an automatic function (as the logo is pretty much mandatory if you want the rom to work on the actual system,) everything out there that already uses WLA-GB puts in the logo manually, and I'd rather not give everyone the inconvenience of looking at a bunch of memory overwrites.

### DIFF
--- a/README
+++ b/README
@@ -135,6 +135,7 @@ Group 2:
 
 [GB ] .CARTRIDGETYPE 1
 [GB ] .COUNTRYCODE 1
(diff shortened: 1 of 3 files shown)


vhelin:
Thanks! :+1:

Issue #33, opened 2015-05-01
Local Labels


Drenn1:
First off I know that if you put an underscore before a label name, it exists only in the given section. I also know about nameless labels. However, I would find it extremely useful to be able to make a local label which will fall out of scope after the next non-local label.

Requesting this because I found myself jealous of the pokered disassembly, which uses rgbds and puts a .done label at the end of every function which needs it. The readability definitely beats unnamed labels.

Drenn1:
Okay so, I kinda-sorta dealt with this myself. I didn't touch wla because I didn't really want to be responsible for bugs and whatnot, plus I wasn't sure how I'd even do it. Instead I wrote a [hackalicious script](https://github.com/Drenn1/ages-disasm/blob/82205762be767148ac4c0923ea86cd3f3cc0be6e/tools/wlaParseLocalLabels.py) which runs through my assembly file. Labels starting with a dot are considered "ultra-local" labels. Sorry if this is unhelpful, but well, I'm the only one who asked for this feature anyway :P

Anyway it looks like this if you're interested:

```
function1:
    ; do stuff
.done:
    ret

function2:
    ; do different stuff
.done:
    ret
```

galaxyhaxz:
May I make a suggestion? If this feature gets added, make it work like the above but use "@" instead of ".", since commands use a period. Also, "@" was used in real life here's a sample from the B.O.B source code (excuse the bad formatting):

	lda	#1
	sta	gettitlemusic
	bra	@gameover

	@notademo:
	lda	lives
	bne	@notgameover
	lda	#1
	sta	gotocontinue
	bra	@gameover

	@notgameover:
	dec	lives
	inc	winflag2

	@gameover:
	stz	demoflag
	jmp	resetloop

Neui:
I would like to prefer to have a "option" with dot because it is faster *(for me)* to write. Maybe if it ends with a `:`, then act as the local lablel?

Example:

	.PRINTT "Now assembling function UnSign"
	UnSign: ; Removes sign in 16-bit A
		PHP
		REP #$40
		BIT #$8000
		BNE .done ; Jumps to UnSign.done
		EOR #$FFFE ; XOR it with FFFE, wich deletes/add the minus sign
	.done: ; Creates UnSign.done or so
		PLP
		RTL

*Ok, to be honest, that might not be a good example because a normal `+` would be enough)*

vhelin:
As long as you guys don't want to make your own fork / branch, WLA DX will
[...]

PR #90, opened 2016-02-15
Child Labels

Drenn1:
This is the next step in my quest to hide all the labels from myself...

Implements #33. From the readme:

```
Labels starting with "@" are considered to be child labels. They can only be
referenced within the scope of their parent labels, unless the full name is
specified. When there is more than one "@", the label is considered to be
a child of a child.

Here are some examples of child labels:

PARENT1:
@CHILD:
@@SUBCHILD

PARENT2:
@CHILD:

This is legal, since each of the @CHILD labels has a different parent. You can
specify a parent to be explicit, like so:

        jr PARENT1@CHILD@SUBCHILD
```

It can go 10 levels deep, although you're likely to hit the name length limit by that point. It might be worth bumping that up.


cr1901:
I REALLY am doing a poor job as a maintainer, and I sincerely apologize. Let's wait until Ville submits his new bug fixes he received via email and I'll see about merging this.

Drenn1:
Just for the record, this doesn't conflict with the "parserUpgrade" branch. There is one merge conflict, but it's just whitespace.

So, you can merge this any time now... =P

vhelin:
Oh man, GitHub doesn't let me merge the pull request, because of that one conflict. Can you Mr. Jones merge this one? :)

Drenn1:
This is actually a different conflict caused by the name-length fix you just merged. Fixed :)

I also recommend, again, increasing MAX_NAME_LENGTH now that you can, since I very frequently ran into that limit, due to the way child labels stack their names together with their parents.

vhelin:
Thank you! Sure, let's increase that MAX_NAME_LENGHT, give me a sec... :)

vhelin:
Thanks!

Just increased MAX_NAME_LENGTH -> 255 - I hope it works all across the
code. :)

Anyway, it seems that at least my CMake here is defaulting to making builds
with WLA_DEBUG and WLALINK_DEBUG defined - are my local sources just
somehow weird or are you seeing this behaviour as well?

I've been coding C# for the past 6+ months so I might be forgetting
something here, but building WLA DX with

  cmake -G "Unix Makefiles"

gives me DEBUG executables... ???

Drenn1:
I'm not getting that behaviour... that's really weird. Did it only started happening after merging this branch? :/

cr1901:
`-DCMAKE_BUILD_TYPE=Debug` on the command-line for Debug binaries, which I think should be the default?

Issue #91, opened 2016-02-18
.DESTINATIONCODE as an alias for .COUNTRYCODE


nicklausw:
I'm basically making this issue as a reminder for myself to implement this already. Technically "destination" makes more sense here than "country" anyway, so.

vhelin:
Looks like this is already in the latest sources. So thanks! :)

PR #93, opened 2016-02-26
DESTINATIONCODE as an alias for COUNTRYCODE #91

nicklausw:
An alias, simple enough.

### DIFF
--- a/README
+++ b/README
@@ -135,6 +135,7 @@ Group 2:
 
 [GB ] .CARTRIDGETYPE 1
 [GB ] .COUNTRYCODE 1
+[GB ] .DESTINATIONCODE 1
 [GB ] .NINTENDOLOGO
 [GB ] .GBHEADER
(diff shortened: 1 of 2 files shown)


vhelin:
Thanks! :)

nicklausw:
I plan on modifying the output so that people don't get confused when they do this:
```
.countrycode
.destinationcode
```
and get an error because one or the other was defined twice. Perhaps `.COUNTRYCODE/DESTINATIONCODE was defined twice` or something like that?

vhelin:
On Sat, Feb 27, 2016 at 9:46 PM, nicklausw <notifications@github.com> wrote:

> I plan on modifying the output so that people don't get confused when they
> do this:
>
> .countrycode
> .destinationcode
>
> and get an error because one or the other was defined twice. Perhaps .COUNTRYCODE/DESTINATIONCODE
> was defined twice or something like that?
>

That (solution) sounds good!

PR #96, opened 2016-03-13
strcaselesscmp conforms with FAILED and SUCCEEDED

nicklausw:
For cleanliness and conformity reasons, strcaselesscmp works like any other WLA function rather than like strcmp.

### DIFF
--- a/pass_1.c
+++ b/pass_1.c
@@ -168,19 +168,19 @@ __far /* put the following big table in the FAR data section */
 int strcaselesscmp(char *s1, char *s2) {
 
   if (s1 == NULL || s2 == NULL)
-    return 0;
+    return FAILED;
 
   while (*s1 != 0) {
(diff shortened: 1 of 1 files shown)


nicklausw:
I'm not completely sure about this to be honest. It's good for strcaselesscmp to act like the function it's based off of (strcmp) but is good for it to work like the rest of the program.

nicklausw:
Update: this, being a big overhaul, doesn't seem to work very well for some directives. Closed 'till whenever.

vhelin:
Sorry for ignoring you guys, I'm travelling in the Asia, have around 80 days left of my journey... Anyway, there's really no need to wrap system functions. :)

nicklausw:
Yeah, that's a good way of putting it. Although in the process of making this PR, I noticed that some directives still use strcmp. Should I go for consistency and make another request for them in specific, or should I just not bother?

Issue #97, opened 2016-03-30
wlalink: cmd arg for silent discard sections


nicklausw:
wlalink's ability to shed unused sections is good, but I don't necessarily want to be reassured every single build that sections are being discarded. I'm thinking "n" can be a command line argument that discards sections, but quiets down on telling you what is being discarded.

vhelin:
The unreferenced section discarder now obeys the -v flag of the linker, which should take care of this, I hope. :)

sverx:
I just stumbled upon this, I didn't realize it had changed.

Honestly I would prefer to have those messages back, but I also see the point in obeying the -v flag... so I wonder if it's possible to send the discard warning messages to _stderr_ in any case, even when `-v` is not specified?

What I mean is that I would love to get the `DISCARD: <section>` messages back but I can't usually be bothered with the other verbose info.

I could also dig a different verbosity level where discard messages are printed and then there's one (or two) line(s) summary at the end such as 
```
ROM: 174339 bytes (66.51%) free of total 262144.
RAM: 2415 bytes (29.48%) free of total 8192.
```
which could be set by a different parameter such as `-v0` or whatever makes more sense... :man_shrugging:

vhelin:
That -v0 things sounds good:

-v0:
Show DISCARD messages

-v1:
Show DISCARD messages
Show ROM & RAM summaries

-v2:
The current -v

?

sverx:
I just got this message even if I'm not using `-v` at all, so there is at least one message (even though it's in wla-z80) that didn't obey the verbose flag.

`DISCARD: HUD.inc: Discarding an empty section "HUD_update".`

vhelin:
Thanks for the info! I'll try to take care of this issue this weekend...

vhelin:
> I just got this message even if I'm not using `-v` at all, so there is at least one message (even though it's in wla-z80) that didn't obey the verbose flag.
> 
> `DISCARD: HUD.inc: Discarding an empty section "HUD_update".`

Ok, the latest commit should also take care of this.

sverx:
It seems to work. Which means that if you want the discard section messages to be printed you now need to turn on verbose, which will print lots of other info too. I guess we need -v1 (and maybe -v2) here too. Not that fundamental, though.

PR #98, opened 2016-03-30
wlalink: silent discard sections with -n

nicklausw:
#97. It gets annoying having tons of messages on the command line saying that sections are discarded, so here's an option for the program to keep quiet.

### DIFF
--- a/README
+++ b/README
@@ -2909,7 +2909,7 @@ After you have produced one or more object files and perhaps some library
 files, you might want to link them together to produce a ROM image / program
 file. "wlalink" is the program you use for that. Here's how you use it:
 
-"wlalink [-divsS]{b/r} [LIBDIR] <LINK FILE> <OUTPUT FILE>"
+"wlalink [-bdnirsSv] [LIBDIR] <LINK FILE> <OUTPUT FILE>"
 
 Choose 'b' for program file or 'r' for ROM image linking.
 
@@ -2980,6 +2980,9 @@ and definitions.
 If flag 'd' is used, WLALINK discards all unreferenced FREE and SEMIFREE
 sections. This way you can link big libraries to your project and WLALINK
 will choose only the used sections, so you won't be linking any dead code/data.
+
+If flag 'n' is used, WLALINK discards sections as if 'd' were used, but does
+not print to the console whenever a section is discarded.
 
 If flag 'i' is given, WLALINK will write list files. Note that you must
 compile the object and library files with -i flag as well. Otherwise WLALINK
--- a/wlalink/discard.c
+++ b/wlalink/discard.c
@@ -15,6 +15,7 @@ extern struct reference *reference_first, *reference_last;
 extern struct section *sec_first, *sec_last;
 extern struct label *labels_first, *labels_last;
 extern struct stack *stacks_first, *stacks_last;
+extern int silent_discard;
 
 
 int discard_unused_sections(void) {
@@ -51,8 +52,9 @@ int discard_unused_sections(void) {
   s = sec_first;
   while (s != NULL) {
     if (s->alive == NO)
-      fprintf(stderr, "DISCARD: %s:%s: Section \"%s\" was discarded.\n",
-	      get_file_name(s->file_id), get_source_file_name(s->file_id, s->file_id_source), s->name);
+      if (silent_discard == OFF)
+        fprintf(stderr, "DISCARD: %s:%s: Section \"%s\" was discarded.\n",
+	        get_file_name(s->file_id), get_source_file_name(s->file_id, s->file_id_source), s->name);
     s = s->next;
   }
 
--- a/wlalink/main.c
+++ b/wlalink/main.c
@@ -41,7 +41,7 @@ unsigned char *rom, *rom_usage, *file_header = NULL, *file_footer = NULL;
 int romsize, rombanks, banksize, verbose_mode = OFF, section_overwrite = OFF, symbol_mode = SYMBOL_MODE_NONE;
 int pc_bank, pc_full, pc_slot, pc_slot_max;
 int file_header_size, file_footer_size, *banksizes = NULL, *bankaddress = NULL;
-int output_mode = OUTPUT_ROM, discard_unreferenced_sections = OFF, use_libdir = NO;
+int output_mode = OUTPUT_ROM, discard_unreferenced_sections = OFF, silent_discard = OFF, use_libdir = NO;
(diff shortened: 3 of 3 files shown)


vhelin:
Good idea, but isn't "-q" usually used to "quiet" things down? :)

nicklausw:
Yeah, but I don't want to completely silence the program, and if I ever did then I'd want -q to be used for that.

vhelin:
How's this usually taken care of?

  -q:d
  -q:discard
  --quiet:discard
  --quiet=discard

??? Using a random letter just to make one option's output to disappear is
not very logical...

nicklausw:
We could have more separate verbose options like -quiet-discard, but this wouldn't work very well with how the program takes input.

bazzinotti:
> Using a random letter just to make one option's output to disappear is
not very logical...

I agree

> How's this usually taken care of?
  -q:d
  -q:discard
  --quiet:discard
  --quiet=discard

Most programs I use in the wild (I'm mostly a OSX / Linux user) will allow a single-letter flag (`-q`) and and expanded version (`--quiet`).

For a flag to accept multiple sub-options, I typically see a comma-dilineated argument. Here's are some examples:

`-q discard,foo`
`--quiet=discard,foo`
`--quiet discard,foo`

> We could have more separate verbose options like -quiet-discard

Comma-separated sub-options can save from the word 'quiet' repeated for every single quiet-option. Since `quiet` sub-options sound like something that could grow, I rest my case.

---

I'm not familiar with WLA-DX's arg parsing impl, but `getopt_long` is a long-proven method to implementing proper argument parsing on *nix. http://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Option-Example.html#Getopt-Long-Option-Example. Their `struct` could be used as an inspiration. I believe there are also `getopt_long` windows ports, but I bet we can simply morph the current parsing code a little bit.

As far as parsing the comma-separated sub-options, there's a neat C function I remember off the top of my head for this: 

```
/* parse single string fields separated by commas */
#include <stdio.h>
#include <string.h>
int main()
{
  char test[] = "discard,foo,bar";
  char sep[] = ",";
  char *word;

  for (word = strtok(test, sep);
       word;
       word = strtok(NULL, sep))
  {
    printf("%s\n", word);
  }
}
```

Anyone experienced in string parsing will probably laugh at the mention of `strtok` because it's implied to understand it long before involving oneself to write something like WLA-DX, am I right? But I realize all kinds of people read here, so I post it succinctly.

vhelin:
On Mon, Apr 18, 2016 at 12:06 AM, Bazz <notifications@github.com> wrote:

> Using a random letter just to make one option's output to disappear is
> not very logical...
>
> I agree
>
> How's this usually taken care of?
> -q:d
> -q:discard
> --quiet:discard
> --quiet=discard
>
> Most programs I use in the wild (I'm mostly a OSX / Linux user) will allow
> a single-letter flag (-q) and and expanded version (--quiet).
>
> For a flag to accept multiple sub-options, I typically see a
> comma-dilineated argument. Here's are some examples:
>
> -q discard,foo
> --quiet=discard,foo
> --quiet discard,foo
>
> We could have more separate verbose options like -quiet-discard
>
> Comma-separated sub-options can save from the word 'quiet' repeated for
> every single quiet-option. Since quiet sub-options sound like something
> that could grow, I rest my case.
>

This sounds good also if it's also the defacto way of doing the silencing,
and yeah, future proof as well.

-q discard,foo
--quiet discard,foo

I'm happy with either one of these (or both). :)



> ------------------------------
>
> I'm not familiar with WLA-DX's arg parsing impl, but getopt_long is a
> long-proven method to implementing proper argument parsing on *nix.
> http://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Option-Example.html#Getopt-Long-Option-Example.
> Their struct could be used as an inspiration. I believe there are also
> getopt_long windows ports, but I bet we can simply morph the current
> parsing code a little bit.
>
> As far as parsing the comma-separated sub-options, there's a neat C
> function I remember off the top of my head for this:
>
> /* parse single string fields separated by commas */
> #include <stdio.h>
> #include <string.h>
> int main()
> {
>   char test[] = "discard,foo,bar";
>   char sep[] = ",";
>   char *word;
>
>   for (word = strtok(test, sep);
>        word;
>        word = strtok(NULL, sep))
>   {
>     printf("%s\n", word);
>   }
> }
>
> Anyone experienced in string parsing will probably laugh at the mention of
> strtok because it's implied to understand it long before involving
> oneself to write something like WLA-DX, am I right? But I realize all kinds
> of people read here, so I post it succinctly.
>

Well, the world is full of functions, can't know them all. :) Anyway, that
looks like a good way of splitting that list into atoms.

nicklausw:
This is outdated, I'll rewrite it later today. How's --quiet-discard sound?

cr1901:
I'd suggest -q, -qq, -qqq, etc personally. Supporting a single long/GNU argument where everything else is a single letter (or arguments chained together) is jarring to me.

vhelin:
Good point!

  -q

We already have that, but could extend it to handle modes?

  ./wla-gb -q -o output.o main.s

... is the old way and should work, plus this new

  ./wla-gb -q the-current-mode-needs-a-name -q discard -q foo -q bar -o
output.o main.s

? Ok? Or the current -q could also be split into sub modes... ?

nicklausw:
I could do that. There'd be two for now, maybe the old one is "output" and then "discard"?

vhelin:
I think I'll look into what the -q does currently first... :)

nicklausw:
Now I know that wlalink had no quiet mode before. I am silly.

So I've added in wlalink's one quiet type, which is discard, and it just looks wrong.

```

WLALINK GB-Z80/Z80/6502/65C02/6510/65816/HUC6280/SPC-700 WLA Macro Assembler Linker v5.8b
Written by Ville Helin in 2000-2008 - In GitHub since 2014: https://github.com/vhelin/wla-dx
USAGE: wlalink [OPTIONS] <LINK FILE> <OUTPUT FILE>

Options:
-b  Program file output
-d  Discard unreferenced sections
-i  Write list files
-r  ROM file output (default)
-s  Write also a NO$GMB symbol file
-S  Write also a WLA symbol file
-v  Verbose messages
-q [QTYP]  Quiet output
    DISCARD  Silent discard unreferenced sections
-L [DIR]  Library directory
```

Is there a better way of describing quiet types properly in the args? This is why I just wanted to add verbose arguments like --quiet-discard.

nicklausw:
...Or since verbose is out of the picture, would just -qd work since there's no other quiet options?

PR #101, opened 2016-04-16
fix modulo errors

nicklausw:
Literally a one-line fix. Can whoever sees this first please merge this?

It fixes the fact that this doesn't work:
`#$00`
but this does:
`$00 + #$00`

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -189,7 +189,7 @@ int stack_calculate(char *in, int *value) {
       in++;
     }
     else if (*in == '#') {
-      if (q == 0) {
+      if (q != 0) {
 	if (input_number_error_msg == YES)
 	  print_error("Syntax error. Invalid use of modulo.\n", ERROR_STC);
 	return FAILED;


bazzinotti:
Pretty vague PR description.. I believe this has macro invocation in mind since we were talking about it in IRC. Are you sure there are no side-effects of reversing the `q` test?

What is that `q` test for anyways (would you mind explaining?)

`$00 + #$00` sounds like something that should never work!

... *sigh* I wish we had a Test Framework so this couldn't be such a guessing game.. just run it against the test cases.. and bam..

nicklausw:
The reason it fixes #52 without breaking anything else is because with regular instructions (lda #$50), the stack calculator isn't used. It only seems to be used with macros. So this shouldn't break anything else as far as I know, but I'll try a few tests regardless.

nicklausw:
Well, my hello world test (https://github.com/nicklausw/hello-nes/tree/master/wla-dx) still builds fine, just with some bank/slot warning that was there anyway.

This works just fine too:
```
.macro load_a
lda \1
.endm

.section "bla"
load_a #$40
.ends
```

There are only so many WLA 65xx projects for me to test on, though...

bazzinotti:
> It only seems to be used with macros

So we need to clearly identify what else the stack calculator is used for then, to be absolutely sure.

What does the stack calculator do? What's the `q` testing?

nicklausw:
It's designed for cases like this:

```
beginning:
.incbin "data"
end:

lda #end-beginning ; stack calculator handles this
```
My guess is that only macros pass on the modulo symbol, which causes an error. Maybe the q test was right, and beginning modulos shouldn't be there at all?

bazzinotti:
@vhelin's response https://github.com/vhelin/wla-dx/issues/52#issuecomment-138098511 is enlightening. And my response to him, as well...

nicklausw:
Probably outdated.

mrehkopf:
The issue persists to date though.

PR #113, opened 2016-06-01
Legacy define, include, library dir support

nicklausw:
These should work okay, so projects with:
```
-Dbla -Ldir -Idir
```
will still assemble.

Spacing in the code went a little everywhere, sorry...


vhelin:
Thanks!

PR #114, opened 2016-06-01
no-library-file macros

nicklausw:
The same code but not as repetitive with making sure libraries aren't getting things they shouldn't.

### DIFF
--- a/pass_1.c
+++ b/pass_1.c
@@ -166,10 +166,12 @@ __far /* put the following big table in the FAR data section */
 
 
 #define no_library_files(name)\
+ do {\
  if (output_format == OUTPUT_LIBRARY) {\
     print_error("Library files don't take " name ".\n", ERROR_DIR);\
     return FAILED;\
-  }
+  }\
+ } while (0)
 
 
 int strcaselesscmp(char *s1, char *s2) {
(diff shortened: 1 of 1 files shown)


cr1901:
This is a good idea. A few style changes: please wrap your macro in a do{}while(0): http://c-faq.com/cpp/multistmt.html, and make sure all directives are qualified with a period in your macro invocation when it matches the old code (your "SLOT" macro invocation is missing the leading period).

nicklausw:
So wait, does a do while statement always loop at least once? I assume so since 0 never equals a non-zero value.

nicklausw:
It works apparently so here you go.

vhelin:
Nice, thanks! :)

nicklausw:
Now I'll just have to think of a bunch of other ways to TRULY shrink pass_1.c's mass...it's not really good that one file in the project is 7000 lines long, but most of it is one function so it can't be helped.

vhelin:
If someone wants to start refactoring, pass_1.c and that one huge function
is a great place to start from. :) But before doing anything major there,
I'd like to see the new branch parserUpgrade assemble and link everybody's
projects first, and after it gets merged to master, then it'd be a good
time to do something else biggish as otherwise we have too many changes in
the repo and too little testing plus all the conflicts... :)

cr1901:
>>So wait, does a do while statement always loop at least once? I assume so since 0 never equals a non-zero value.

Yes. do{}while(0) is a C idiom to prevent multi-line macros from causing errors when used in certain contexts, such as if-else.

PR #119, opened 2016-07-16
Add options and install to CMakeLists

Neui:
Convert the "options" (that you would have to uncomment) to the CMake built-in options. You'll use it like `cmake -DWLALINK_DEBUG=ON ..`.

Also, add an install-target, so you can use `make install`. Tested on Ubuntu and MinGW64 (MSys2).


vhelin:
Nice, thank you!

PR #120, opened 2016-07-17
Bugfixes to wlad

Drenn1:
These fixes are actually rather old. For some reason I never submitted them before?

Anyway, there were some bugs from the last time I worked on wlad. The addresses would get misaligned when dumping from multiple banks and it would basically give incorrect output. This also removes the "debug" opcode so it works with more recent wla versions.


vhelin:
Thanks!

PR #122, opened 2016-07-18
Updated .gitignore and .gitattributes

Neui:
Updated `.gitignore` to ignore swap-files <sub>(i have vim opened when commiting and git asks me if i want to commit these swap-files..)</sub> and ignore build-directories (usually they are called `<something>build/`, let's ignore them too). Also, some things were "duplicated". Removed these too. (`dir/` ignores `dir` and it's contents, no need for `dir/*`)

Also, add a [`.gitattributes`](https://git-scm.com/docs/gitattributes) file. With `* text`, we say that text-files should be normalized to LF when commiting. This should ensure that the repository is consistent with the line endings. Also, `*.sh` are forced to be LF because they're usually from unix (and they use LF) where `.bat` have CRLF (Windows, you know).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,39 @@
+
+*       text=auto
+*.txt   text
+*.rst   text
+*.c     text
+*.h     text
+*.md    text
+*.sh    text eol=lf
+*.bat   text eol=crlf
+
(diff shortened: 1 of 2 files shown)


vhelin:
Thanks again! :)

PR #124, opened 2016-09-22
Remove the need for a linkfile

galaxyhaxz:
The new command line works as follows:
wlalink [OPTIONS] \<OBJECT FILES\> \<OUTPUT FILE\>
Libraries are specified just like in ANSI-C. That is, -llibname.
Note that because you can no longer specify the library bank/slot, it defaults to bank 0 slot 0. If that is occupied by rom data, you must write the library's section as SUPERFREE to allow it to be placed in the first free section. The working directory is first searched, then the lib directory. In each case ".lib" will be appended if the library is not found. This lets you specify custom extensions, or use no extension to save time :D

Examples:
wlalink -v -L./lib -ldebug -lc -lm main.o graphics.o sprites.o MyGame.nes
wlalink -S -v -Lsdk/lib -ldrawSprites.lib -ltest.obj game.o Output.bin

TODO:
- still need to remove any existing linkfile/header/footer code. or add commands such as -h "header.bin" -f "footer.bin"
- more testing, I only tested it with my project and a sample library with no problems.
- update readme
- also the parse_flags needs a redundancy check

### DIFF
--- a/wlalink/files.c
+++ b/wlalink/files.c
@@ -14,261 +14,67 @@
 extern struct object_file *obj_first, *obj_last, *obj_tmp;
 extern struct label *labels_first, *labels_last;
 extern unsigned char *file_header, *file_footer;
-extern int file_header_size, file_footer_size, use_libdir;
+extern int file_header_size, file_footer_size, use_libdir, total_flags;
 char file_name_error[] = "???";
 extern char ext_libdir[MAX_NAME_LENGTH];
 
 
 
-int load_files(char *argv[], int argc) {
+int load_objects(char *argv[], int argc) {
 
-  int state = STATE_NONE, i, x, line, bank, slot, base, bank_defined, slot_defined, base_defined, n;
-  char tmp[1024], token[1024], tmp_token[MAX_NAME_LENGTH];
-  struct label *l;
-  FILE *fop, *f;
+  int i;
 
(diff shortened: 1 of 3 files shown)


nicklausw:
Nice, looks good.

Is anyone here to merge it though?

vhelin:
I'm here, but drunk atm, too many beers, and working on a project with
pretty tight deadline coming up soon... I could merge this, though, like
mostly blind. :)

vhelin:
Hey,

"Note that because you can no longer specify the library bank/slot, it defaults to bank 0 slot 0."

Will this mean also that all existing projects will fail with this new version of wlalink? If yes, could it by default work as previously, but in addition to that could be instructed this new way? I'd love to keep backwards compatibility here...

galaxyhaxz:
Well with the absence of a linkfile It isn't possible to specify the library's location so it has to work this way. All I had to do for my project was type SUPERFREE for the library. If you have to specify the lib's slot or base then currently your out of luck. I think it shouldn't be too much work to modify the section parser to treat all libraries as SUPERFREE and find the first free location regardless of slot. Then a .base command could be added for libraries.

galaxyhaxz:
Closing for now. I'm going to fix some bugs before I reopen this one.

PR #126, opened 2016-10-11
Add a .ROMGBCONLY directive

Drenn1:
The .ROMGBC directive writes $80 to $143 in the cartridge header, contrary to what the readme said. This indicates that it's a "dual-mode" game, which can run in either GBC or DMG mode.

The new .ROMGBCONLY directive writes $c0 to $143, indicating that it can ONLY run in gbc mode.

I also added this to the .gbheader structure, although I'm not entirely convinced that other aspects of the gbheader are working correctly... this commit also fixes "nintendologo" so that it works with .gbheaders, though there may be other issues.

### DIFF
--- a/README
+++ b/README
@@ -155,6 +155,7 @@ Group 2:
 [GB ] .RAMSIZE 0
 [GB ] .ROMDMG
 [GB ] .ROMGBC
+[GB ] .ROMGBCONLY
 [GB ] .ROMSGB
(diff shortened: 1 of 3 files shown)


vhelin:
Nice, thanks! 👍

nicklausw:
Could be wrong, but I recall designing gbheader so that nintendologo is placed automatically.

PR #133, opened 2017-01-05
Check for null byte when trimming linkfile line

### LINKED ISSUE #79: wlalink crashes
[linkerbug.zip](https://github.com/vhelin/wla-dx/files/92824/linkerbug.zip)
I'm on the newest version of wla-dx. Wlalink crashes when I link objects and libraries. I have attached a minimal project that demonstrates this error. I have an older (before GitHub) version of wla-dx, and if I use the wla-z80 and wlalink from this version on the same files, no error, and everything links just fine.

Gekkio:
If EOF happens before fgets reaches a newline, the resulting buffer
shouldn't be read past the automatically added null byte.

Most likely fixes #79

### DIFF
--- a/wlalink/files.c
+++ b/wlalink/files.c
@@ -43,7 +43,7 @@ int load_files(char *argv[], int argc) {
       continue;
 
     /* remove garbage from the end */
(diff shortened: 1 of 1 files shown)


vhelin:
Looks good, kiitoksia! 👍

PR #135, opened 2017-01-10
Fix negative range for 8-bit two's complement

VectorKiwi:
Various checks were considering -128 to be out of range for a signed 8-bit value.


vhelin:
Great, thank you! 👍

Issue #136, opened 2017-02-07
_sizeof_[label name]


maxim-zhao:
It's be nice if there was an auto-generated symbol for each named label which was the count in bytes from it to the next such label. This would make it a lot easier to capture the lengths of data blobs, for example:
```
label:
.asc "Some message here"
.define _sizeof_label next_label-label ; automate this
next_label:
```
Typically I find that the .define doesn't work nicely because of the lexical positioning typically being after where I want to use it anyway.

vhelin:
Ok, WLALINK now generates these definitions. I hope it works 100%. :)

maxim-zhao:
Awesome! I'll give it a spin later.

Ramsis-SNES:
This is indeed a cool feature, but unfortunately I can't get it to work. If I replace e.g. [this line](https://github.com/Ramsis-SNES/furryrpg/blob/master/furryrpg_text.inc.asm#L289):

`cmp	#(SRC_DiagPointerEng_END-SRC_DiagPointerEng)/2`

with this:

`cmp.w	_sizeof_SRC_DiagPointerEng/2`

then I get a value of zero instead of $BC. What am I doing wrong?

Also, it would be nice if the current size of the Accumulator would be acknowledged so `.w` could be omitted. :)

vhelin:
Can you post here the code? E.g. The section that contains both of the
labels.

In the case of _sizeof_ defines, the linker creates then, and that is too
late for size deduction.

Ramsis-SNES:
Sure, this is the file with the labels: https://github.com/Ramsis-SNES/furryrpg/blob/master/data/ptr_dialog_eng.inc.asm

The two labels are at the beginning and end of the file, respectively. (I didn't remove the `..._END` label when testing `_sizeof_`, BTW.)

For further reference, this is the main assembly file which contains most (if not all) `.INCLUDE`s: https://github.com/Ramsis-SNES/furryrpg/blob/master/furryrpg.asm

vhelin:
Thanks!

Looking at the first file I have no idea why the distance between the
labels is calculated incorrectly as the sources are as simple as it gets.
I'll try to debug this later today.

vhelin:
I wrote a similar testcase, but there the size of _sizeof_ definition was correct. If you make WLALINK to write out a symbol file, what's the value of _sizeof_SRC_DiagPointerEng there?

Ramsis-SNES:
[...]

PR #145, opened 2017-07-03
Make _sizeof labels choose the label hashmaps more wisely.

Drenn1:
This should fix issues with multiple labels of the same name across different sections. I refactored out the code that decides which hashmap a label goes into, to use for the _sizeof labels.

_sizeof labels will inherit the "locality" of the base label they came from. So, despite starting with an underscore, they won't always be local. (And they won't always be global like before.) I put a few tests in my "namespaces" example.

This doesn't do anything with the "@" labels, which as @maxim-zhao mentioned, maybe should behave a bit differently, though I don't really care much about that... =P

This is a followup to #136.

### DIFF
--- a/README
+++ b/README
@@ -2717,13 +2717,25 @@ calculations and sees only the preprocessed output of WLA):
                 ; a macro).
 
 WLALINK will also generate _sizeof_[label] defines that measure the distance
-between two consecutive labels. Here is an example:
+between two consecutive labels. These labels have the same scope as the labels
+they describe. Here is an example:
 
 Label1:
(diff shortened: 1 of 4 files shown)


Drenn1:
Nevermind, I guess I do care a little... =P

You can take or leave that last commit. That tells it to skip over child labels when calculating the size.

maxim-zhao:
Well, it conflicts nicely with my attempt :smile: which I haven't got around to making a PR for yet. I do have a codebase very reliant on it to test with, though.

vhelin:
Thanks!

PR #149, opened 2017-07-09
Off-by-1 error with nintendo logo

Drenn1:
I was getting a "data being overwritten" warning when I tried defining the "cartridge title" manually, which comes directly after the nintendo logo...

A harmless bug, at any rate.


vhelin:
Thanks!

Issue #30, opened 2015-02-19
Can't use .instanceof in enum


maxim-zhao:
Docs say:

> DB, DW, DS, DSB, DSW and INSTANCEOF can also be in lowercase. You
> can also use a dotted version of the symbols, but it doesn't advance
> the memory address. Here's an exmple:

(Note, typo at the end)
But:

```
.struct XY
  y db
  x db
.endst
.enum 0
  foo .instanceof XY
  bar dsb 100
.ende
```

```
DIRECTIVE_ERROR: Unexpected symbol "bar" in .ENUM.
ERROR: Couldn't parse "bar".
```

In fact, anything that comes after the struct name is rejected.

---

Maybe there's a better way to achieve what I want to do here? Namely:
1. I have Z80 code using the index registers, so it's super useful to use the struct member offset definitions (e.g. XY.x above) for the relative distances. Thus I want to use structs.
2. I also have cases where two totally separate parts of the code (title screen and in-game) want to make use of the same RAM locations for unrelated purposes. 

If I have .instanceof then I can make suitable structs to overlay the same part of RAM (much like a C union of structs) and put them in the same containing .enum. If I don't then I guess my workaround is to split my .enum up, so I can have two enumerating through the relevant bytes, but then I'm required to manage the start addresses to avoid unintended clashes - which is the problem .enum solves in the first place.

NicklausW:
Wait, what happens if you use instanceof instead of .instanceof?

maxim-zhao:
Well, it works - my point is, I want it to overlap with the "bar" array. How about a better example...
```
.struct InGameData
  foo db
  bar dsb 20
  baz dw
.ends
.struct TitleScreenData
  buffer dsb 16
  counter dw
.ends
.enum $c000
  ; Some items before the overlapped bit. Let's assume they're used by all the code.
  var1 db
  var2 dw
  ; I want to put an instance of each struct here, overlapping
  RAM_TitleScreenData .instanceof TitleScreenData
  RAM_InGameData instanceof InGameData
  ; more stuff here
  var3 dw
.ende
```
Trying to define all the relevant items without structs would be tough due to the way they use non-coincident chunks of RAM, and I want a struct for one or more of them anyway. Having .instanceof still requires me to select the largest struct to advance the enum (if I want to guarantee that following items don't get overlapped).

[...]

PR #150, opened 2017-07-11
Allow the use of ".IF" directives within enums and ramsections

Drenn1:
As usual, I did this because I desperately needed this feature. I'm integrating two games, Oracle of Ages and Seasons into the same codebase, and I absolutely need to define RAM conditionally.

This was easier to do than I thought it would be. Enum and ramsection code has been moved into a separate function, moved outside of the "parse_directive" function to be part of the "evaluate_token" loop. This has the added benefit of refactoring the very similar enum and ramsection code to be together.

The "IF" directives have been moved into the "parse_if_directive" function, which is like "parse_directive" but only for those IF directives. It's reused by the enum/ramsection code.

Looking back, I probably could have done this without refactoring the enum and ramsection stuff out of parse_directive, but... eh.

In addition, I noticed that the code was checking for literally any string starting with a dot, and treating that as a "non-advancing define". That... felt really wrong. I've changed it to only check for ".db" and ".dw". But ".dsb", ".dsw", and ".instanceof" weren't working anyway because it wasn't checking for the parameters that come after. So, bug #30 still applies.

I did this without consulting anyone, so please point out if I did this in a non-preferred way or something.

### DIFF
--- a/README
+++ b/README
@@ -1872,6 +1872,8 @@ And this is what is generated:
 This way you can generate a 16-bit variable address along with pointers
 to its parts.
 
+You can use all ".IF" and ".ELSE" directives within an enum.
+
 If you want more flexible variable positioning, take a look at
 .RAMSECTIONs.
 
@@ -2328,6 +2330,8 @@ If no other RAM sections are used, then this is what you will get:
 
 BANK in .RAMSECTION is optional so you can leave it away if you
 don't switch RAM banks, or the target doesn't have them.
+
+You can use all ".IF" and ".ELSE" directives within an enum.
 
 This is not a compulsory directive.
 
--- a/pass_1.c
+++ b/pass_1.c
@@ -125,6 +125,10 @@ int xbit_size = 0;
 int accu_size = 8, index_size = 8;
 #endif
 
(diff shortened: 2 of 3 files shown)


vhelin:
Thank you! The diff for pass_1.c was long and looked weird here on GitHub, didn't really understand it all, but lets hope it works. :)

Kroc:
Excellent! Round-of-applause

nicklausw:
While we're at it, couldn't all directives that have their own parser be rewritten like this? As in .GBHEADER and things like that.

vhelin:
It should be possible, but who is going to do all the work? :)

nicklausw:
I could try it, but it honestly wouldn't be a high priority of mine because I barely use WLA anymore. It sounds like a lot of "template work", just looking at the enum parser and going from there with everything else, which I don't mind doing.

Drenn1:
It should actually be easy. I didn't need to reorganize it as much as I did. Most of that came from a primal urge to refactor.

It should just involve calling parse_if_directive from within each directive... I'll see about doing that, assuming you haven't already started.

PR #152, opened 2017-07-11
Allow .IF directives on all other directives with custom parsers

Drenn1:
Taking @nicklausw's suggestion from #150.

Affected directives:
* GBHEADER
* SMSHEADER
* SNESEMUVECTOR
* SNESHEADER
* SNESNATIVEVECTOR
* ASCIITABLE
* MEMORYMAP
* ROMBANKMAP
* STRUCT

I haven't tested all of these, but most of the examples still work (see below...). Since it was a simple copy/paste job, I'd be surprised if any of them broke, though.

I remembered when I was going through this that STRUCT is much the same as ENUM and RAMSECTION, so I applied the same change related to dotted labels. I guess ideally I'd refactor that too, but... eh.

Another note: the enum/ramsection/struct change to dotted opcodes has caused an error in spc-700/linker_test. You could argue that the test case is broken, since it uses the enum in a way not specified in the readme...

### DIFF
--- a/README
+++ b/README
@@ -1598,8 +1598,9 @@ The following operators are supported:
 == - equals to
 != - doesn't equal to
 
-All IF (yes, including .IFDEF, .IFNDEF, etc) directives can be
-nested.
+All IF directives (yes, including .IFDEF, .IFNDEF, etc) can be
+nested. They can also be used within ENUMs, RAMSECTIONs, STRUCTs,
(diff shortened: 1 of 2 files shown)


vhelin:
Thank you!

vhelin:
I fixed spc-700/linker_test, works now.

PR #157, opened 2017-08-05
Speed up searching for anonymous labels

Drenn1:
All this changes is that the "get_closest_anonymous_label" function in wlalink will only search anonymous labels, not all labels. This reduced wlalink's runtime from 18 seconds to 3 seconds on my laptop.

I was originally planning to do a binary search, but I'm satisfied with this improvement for now. There's no real reason the anonymous label list needs to be sorted, though, since I'm not doing that.


vhelin:
Thanks! And sorry, I somehow missed this PR before...

Drenn1:
I had completely forgotten about this, lol.

PR #161, opened 2017-10-26
Universal base

SolraBizna:
This PR allows `.BASE` directives to be used on all targets, not just 65816. I did small-scale tests of all the edge cases I could think of. I also corrected the documentation regarding what `.BASE` actually does.

This feature is useful to me because I'm working on a project that uses two separate ROM chips, each mapped to a different range of hardware banks.

(This PR was ready weeks ago, but I had a poorly timed hardware failure.)


vhelin:
Thanks!

PR #154, opened 2017-07-24
More documentation sphinx fixes (1.2.3 support)

Neui:
Continuation of #153. Cygwin still ships sphinx 1.2.3 in their repositories. This PR tries to make it work on that (and newer) version. Please test this PR before merging! I don't think you really want so many "useless" PRs when it can be done just in one. ;)

(See this [comment's last paragraph](https://github.com/vhelin/wla-dx/pull/153#issue-245133018), be sure to run `cmake .` before so you get the newest Makefiles)

PR #163, opened 2017-12-24
Some documentation fixes (incl. removing wlad)

Neui:
I had these changes for quite for some time while. Now I rebased it and made a PR.

It mostly fixes some formatting issues and also removes anything about wlad, because wlad got removed, so it isn't needed anymore. Also, I've made the singletext-extension function with more sphinx versions, [because it was reported it didn't work on a certain version](https://github.com/vhelin/wla-dx/pull/154#issuecomment-317639537). Just in case. Can't remember I've tested that version.


vhelin:
Nice, thank you! 👍

Issue #166, opened 2018-02-07
Getting bank number with ":CADDR" is broken


Drenn1:
Something broke after the commits relating to the ".BASE" directive (after 9841ae688c8274d150a7bff2a1c385d4f058d2df). I have a line:

```
ld a,:CADDR
```

and it seems to be giving me an undefined value - that is, it's different every time, whether or not the ".BASE" directive was used. Any idea what may be causing this, @SolraBizna?

Drenn1:
If I go to wlalink/write.c and hardcode the base number to "0" for the CADDR case, the problem is fixed. However I suspect it should use the value of the last ".BASE" directive instead. This is a weird case since it's a reference that doesn't have a corresponding label anywhere, so the value of ".BASE" for it doesn't seem to be stored anywhere.

Drenn1:
Here's a sample project. I wasn't getting the undefined behaviour at first, but that appeared after adding another bank reference below it.

[wla_base_bug.zip](https://github.com/vhelin/wla-dx/files/1704341/wla_base_bug.zip)

Also, I found a bonus bug for you. I noticed that "base" is still wrapped in a W65816 define under the "stack" struct in defines.h. This means arithmetic won't be handled properly. I added a line for that in the test project above.

SolraBizna:
Valgrind confirms it. When `fix_references` creates a fake label for `CADDR` to link against, it doesn't initialize `base`. I'll fix that and the stack issue now, and while I'm at it I'll look at every W65816 case in the whole codebase to see if they should be promoted to general cases.

SolraBizna:
Your test case was very helpful. Thanks.

#167 fixes all the remaining oddness I could find. As I mention in the description there, I couldn't find a better way to fix this particular bug than setting `base` to 0 when referencing `CADDR`. A more correct fix would require breaking a rather non-future-proof on-disk format. For now, if anyone wants to use a non-zero `.BASE` *and* wants a correct `:CADDR`, they can address their hate mail to me. :)

vhelin:
On-disk object/library file format can be broken, if you ask me, if it makes a bug fix or a nice feature possible; it's quite easy and fast to recompile all assembly sources, so there are no big reasons why not to change the file formats...

vhelin:
[...]

PR #167, opened 2018-02-07
Respect `.BASE` in more places, partial fix for #166

SolraBizna:
24-bit label math is now (theoretically) possible outside 65816. Overriding the base for library sections is now (theoretically) possible outside 65816. `:CADDR` now gives sensible results again, instead of adding an uninitialized value. (It used to give nonsense results, but only on 65816; it was my previous universal-base efforts that expanded that bug to the other backends.)

`CADDR` still does not respect `.BASE`, but I don't see a simple way to fully handle that without breaking an on-disk format. I don't know what this project's conventions are regarding format stability, but I do know that the current behavior will only cause problems for people using `CADDR` **with** `.BASE`. At least it will now give sensible, if non-`.BASE`d, values for `:CADDR` instead of random ones...

### DIFF
--- a/defines.h
+++ b/defines.h
@@ -89,7 +89,7 @@
 /* O - origin              */
 /* o - absolute origin     */
 /* B - ROM bank            */
-/* b - BASE (65816)        */
+/* b - BASE                */
 /* L - label               */
 /* r - 16-bit reference    */
 /* R - 8-bit pc ref        */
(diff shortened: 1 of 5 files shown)


SolraBizna:
I should also mention that there's a workaround for getting the bank number of the current code even with a non-zero `.BASE`. Instead of:
```
foo:
    ; ...
    ld a,:CADDR
```
one can do:
```
foo:
    ; ...
    ld a,:foo
```

vhelin:
Thanks!

PR #171, opened 2018-04-03
Error messages compatible with emacs hyperlinks

fabiodl:
Added space to make error messages compatible with emacs compilation output links


vhelin:
Thanks!

PR #174, opened 2018-04-21
Unions

Drenn1:
Another feature poached from rgbds: https://github.com/rednex/rgbds/pull/193

Somewhat analagous to a C union, this allows you to reuse memory in an enum, ramsection, or struct. From the documentation:

```
    .ENUM $C000
        .UNION
            pos_lowbyte:  db
            pos_highbyte: db
            extra_word:   dw
        .NEXTU
            pos:          dw
        .ENDU
        after: db
    .ENDE

This example is equivalent to:

    .DEFINE pos_lowbyte  $c000
    .DEFINE pos_highbyte $c001
    .DEFINE extra_word   $c002
    .DEFINE pos          $c000
    .DEFINE after        $c004
```

This involved refactoring enums, ramsections, and structs especially. I'd started on this before when I allowed ifdefs in them; now it's fully refactored. Enums and ramsections now make temporary structs, so all 3 of those things share the same code.

You can also add a name as an argument to ".UNION" or ".NEXTU" to put it into something similar to a namespace. You know how much I like my namespaces.

This isn't finished yet.

- [x] Need to write tests (in particular DESC enums are probably broken right now)
- [ ] Need to fix .DSTRUCT. Old syntax doesn't really make sense anymore. I should be able to fix it when unions aren't in use. But when they are, it will need entirely new syntax. Perhaps something like this?
```
.STRUCT MyStruct
  .UNION
    var1: db
    var2: db
  .NEXTU
    var3: dw
  .ENDU
.ENDST

.DSTRUCT MyStruct VALUES
  var1 $13
  var2 $37
.ENDST
```
Note the "VALUES" keyword... my idea is to use "VALUES" to indicate the new syntax, and "DATA" (or nothing) to indicate the old syntax.

Another note: I feel like ".db" and ".dw" should be deprecated, if that means anything. Well, perhaps they're still sometimes useful, but those two commands are identical, and something like ".none" might be more descriptive. Well, it's probably not worth trying to change at this point...

I'd appreciate feedback on how this is implemented (even if just a thumbs up emoji).


cr1901:
Great idea for structs :)! But in the context of enums, I believe we've always supported having two variables share the same memory address using the `.DW` syntax to not increment the current address in the enum?

Drenn1:
That's always been... messy. I made this because I wasn't satisfied with that.

Observe this example piece of code which reuses memory between two different game menus.

With unions:

```
.union fileSelect

	mode: ; $cbb3
		db
	mode2: ; $cbb4
		db
	selectedFile: ; $cbb5
		db

.nextu mapMenu

	mode: ; $cbb3
		db
	cursorPos: ; $cbb4
		db
	currentRoom: ; $cbb5
		db
.endu
```

Without unions:

```
wFileSelect.mode:
	.db
wMapMenu.mode:
	.db
wTmpcbb3: ; $cbb3
	db

wFileSelect.mode2:
	.db
wMapMenu.cursorPos:
	.db
wTmpcbb4: ; $cbb4
	db

wFileSelect.selectedFile:
	.db
wMapMenu.currentRoom:
	.db
wTmpcbb5: ; $cbb5
	db


```

It's a lot neater with unions; everything related to a particular menu is in one place, instead of being scattered all around. And in my actual project, I have something like 7 menus that reuse memory, each with about 16 variables... you can imagine, it's rather annoying to wade through 7 menus worth of variables when I only want to see one.

And this gets even messier when you have variables of different sizes, ie. words and bytes. You need to order them in a very specific way. Not to mention, it was impossible to use instanceof like this. I looked into supporting dotted instanceof, but I just couldn't think of a sane way to do it. If it was supported, you would need to know exactly how big the struct is so that you would know to skip that amount of memory when your shared memory section is done.

Unions guarantee that a set amount of memory is allocated for all entries in the union, so there's no chance of memory accidentally "leaking" into variables that aren't in the union.

Drenn1:
I suppose the key distinction here is that this is more than just "not incrementing" the current address, this is "going back" to a previous address. It can make things a lot neater.

vhelin:
This .UNION feature looks pretty good! :) The code also looks quite nice, but also introduced some questions:

- should some of the variables be initialized to 0 or NULL (e.g. in pass_1.c: struct union_stack *union_stack)?
- in pass_1.c:enum_add_struct_fields(), should tmp[]'s size be at least MAX_NAME_LENGTH*2+1, possibly even more?
- ... same with union_basename[] later in the same function
- ... in the same function there are malloc()s without NULL checks

Anyway, 👍 😀

Drenn1:
union_stack is initialized to null when an enum/ramsection/struct is entered, which should be enough, but I guess it wouldn't hurt anyway?

I'll address the rest of those, thanks =)

vhelin:
Thanks! Looking forward to this. 😄

Drenn1:
OK, a bunch of stuff today...

I ran into some issues with _sizeof. In particular, since the linker was trying to calculate it based on distance between labels, this resulted a lot of union fields being set to 0. I changed it so that the assembler performs the sizeof calculations for ramsections instead of the linker. I also extended that to enums and structs. In my opinion it's a lot more well-defined this way, anyway, since it's not based on the distance between 2 labels anymore. (Though that's still the case for non-RAM sections.)

I made ramsections, enums, etc. reject unrecognized directives (things that start with '.'). Since they're being used more and more within those sections, I think it makes sense. However this could cause compatibility issues if someone has a name starting with a dot...

I fixed .DSTRUCT, but haven't done a reimplementation to support unions yet. Not sure when I might get to that...

I put a verified in my new test (and one of the 6502 tests) to check that the output is as expected. For the new test, it's actually based on the sym file, not the rom file it produces.

As always, scrutiny is appreciated. :)

Drenn1:
This has long since become unmergeable. My codebase still relies on it, though, so I'm going to redo this. Let me know if you have any comments about how the feature should be implemented.

Drenn1:
I opened a new PR for this, #243.

PR #177, opened 2018-04-30
Fixes typo in .ASCIITABLE docs

bfloch:
> A is equal to 0, B is equal to **0**, C is equal to 2

should be:

> A is equal to 0, B is equal to **1**, C is equal to 2


vhelin:
Thanks!

bfloch:
Thanks for the merge.

Can you glance over the ASCII table sample. I was kind of confused that it defines "!" as 90 but later claims "ABZ"<-- with Z being 90.

I can not write a PR as I don't know what the values were supposed to mean but it looks fishy. Or am I missing something?

PR #164, opened 2017-12-30
[WIP] Better building and such

Neui:
#### ***This is still in WIP (work in progress), so don't merge yet!***

**Note: I am currently separating some changes that I made in this PR into separate PRs so they can already go upstream. This PR will become a PR about adding testing**

This is what I've worked on for some time. It makes the build system better, makes giving versions easier and integrated tests (primary from the examples folder, which was already used as tests).

Also, since this is WIP, I am going to rebased and squash a lot, so some dates will be wrong. I sometimes work on "all" things at once.

- [X] ~~Improved CMake build system~~ See PR #184 
  * [ ] Better version using `git describe` (number of commits since latest annotated tag) (Not satisfied)
    * (Note that the latest tag `v9.6` is a lightweight tag, not an annotated tag)
    * Not happy what version format I should use for dev builds; `{git-describe}`? `{git-describe}-{branch}`?
  * Cross-compiling support
  * Less repetition
- [ ] Tests
  * 2 example tests included that can be fixed in a later PR
  * Tests with libraries are missing
  * Using some simple file checks it builds and checks the output of the commands and (optionally) compares with an `expected.bin`. (Partly #165, since there is no test with such file yet)
  * No negative tests
  * Testing over CTest (/CMake)
    * No support for Makefile-based tests... yet
  * Only uses the binaries wla-* and wlalink
- [ ] Continuous Integration (CI)
  * So a test-build and the tests are executed on e. g. every pull request
  * Need to test when told to build on a tag
  * Travis CI [Travis Build Status](https://travis-ci.org/Neui/wla-dx)
  * Gitlab CI [Gitlab CI Build Status](https://gitlab.com/Neui/wla-dx/pipelines) (not used here, but I did it anyway)
  * AppVeyor AppVeyor Build Status
- [X] ~~[Read the Docs](https://wla-dx-neui.readthedocs.io/en/better-building-and-such/wla-dx.html) support~~ See PR #193 
- [ ] POSIX Makefiles
  * A script generates the makefile(s), but the generated one will be shipped with the git repo for convenience
  * No tests (yet)
- [X] Better `README.md` and similar files
  * Also includes an IRC channel to discuss, but I marked it as 'unofficial', for now

PR #180, opened 2018-05-26
Don't crash when trying to print ops in DEBUG binary

Neui:
> Don't use variadic arguments twice without resetting inbetween.

Was annoying when the DEBUG build crashed while testing in PR #164. I somehow forgot to include in PR #179. Also, I'm unsure if the `ttt` buffer has enough space, but since it is DEBUG only, I don't really care for now.


vhelin:
Thanks! I'll increase the ttt buffer size, just to be sure.

PR #181, opened 2018-06-14
Better git

Neui:
This PR makes some .git-files better. The biggest issue I had was that the 2 files in the `historical` folder would always be marked as being modified and thus some git commands refused to work (e. g. `git rebase`).

I would actually prefer the `historical`-folder being deleted since this is git, you can revert to any version you like. (And/or create a new branch that still has the folder for reference) Maybe I'll do a PR about it, but not now.


cr1901:
I occasionally still use the msdos port (some dev tools only exist for DOS), so is it possible we just update the gitignore _without_ removing the Makefile (to symbolically represent that the port still works just not a priority)?

Neui:
`msdos.bat` was never removed in the first place (neither deleting it or putting it into `.gitignore`).
(I had `historical/msdos.bat` explicit in `.gitattributes` which you might have confused with `.gitignore`)

Anyway, I updated it again to make the `historical/wlad_new/makefile.win32` and `historical/wlad_new/makefile.win32.orig` to be CRLF (when checking in, like it was before) because of some misunderstanding between me and the git docs.

Note that the files would now have LF inside the internal tree because git prefers it. When they get put into the working directory, they would be converted into CRLF, so there should essential be no change.

vhelin:
Thanks! But let's keep historical folder and not delete it, just for convenience. :)

PR #183, opened 2018-06-26
Allow smaller background

maxim-zhao:
When patching a ROM it is not unusual to also "expand" it. This change makes `.background` happy to have the input file be smaller than the output, and leaves the extra space marked as unused.


vhelin:
Thanks!

PR #201, opened 2018-10-06
generate_tmp_files a bit "renovated"

Neui:
* More C like with pointers to save to (but still saves to a global variable)
* Since it seems the 2nd temp file name isn't even being used, this has been removed
* Windows now uses the `.wla%PID%a` format just as the documentation promises
* Special check if the `sprintf` call overflows the buffer, just in case

This was relatively something simple to do after the bit of CMake and the debugging of the macro stack growing thing. Maybe it's better to store the temp file to their respective location (`/tmp` on unix-like, [`GetTempPath`](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-gettemppatha) on Windows, current working directory on others).

### DIFF
--- a/doc/tempfiles.rst
+++ b/doc/tempfiles.rst
@@ -1,20 +1,10 @@
 Temporary Files
 ===============
 
-Note that WLA will generate two temporary files while it works. Both files
-are placed into the current working directory.
-
-======== ============== ==============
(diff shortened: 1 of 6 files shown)


vhelin:
Thanks!

Issue #191, opened 2018-07-27
Add address-to-line mapping and source file list to Symbols file


CypherSignal:
In an effort to try and improve the debugging experience of homebrew ROM development, I have been working on some changes to wla-dx to add some additional information to the .sym file output by the linker: an listing of the address-to-line mapping, plus a list of source-files that were included during obj compilation. The majority of the changes that I have so far rely on the existing list cmds that are generated in the obj files, so it's mostly outputting data already collected. There are some additional things saved and calculated such as source-file checksum at time of obj compilation, or the base of a given section's bank.

For example, the following is added to the .sym file for a [simple ROM from Neviksti's starter kit](http://wiki.nesdev.com/w/index.php/Nesdev#SNES), after the `[labels]` and `[definitions]` sections : 
```
[source-files]
0001 a22631df show.asm 
0002 cd810adf show.inc 
0003 07116f72 LoadGraphics.asm 
0004 3117b7df InitSNES.asm 
0005 c18e5c0e 2input.asm 

[rom-file checksum]
a3afd3d7

[addr-to-line-mapping]
00:8383 0002:00000059
00:8384 0002:0000005c
00:8386 0002:0000005d
00:8387 0002:0000005e
00:8388 0002:00000060
00:838a 0002:00000061
00:838d 0002:00000063
00:838e 0002:00000064
00:838f 0002:00000065
00:8311 0003:00000049
00:8312 0003:0000004a
00:8313 0003:0000004b
00:8314 0003:0000004c
00:8315 0003:0000004d
00:8316 0003:0000004f
00:8318 0003:00000052
00:831b 0003:00000053
00:831e 0003:00000054
00:8321 0003:00000055
00:8324 0003:00000057
00:8326 0003:00000058
...
```

...et cetera. A full gist of the full .sym file can be seen here: https://gist.github.com/CypherSignal/fb23d610f2f835caf6bbc7ba426be9ef

When acknowledged by a corresponding emulator that loads up the addr2line map and referenced source files, this can be used to display the source assembly of a disassembled ROM, or perform other operations using the source file as a reference, such as add breakpoints by referring to a line in a source file, as opposed to solely relying on addresses. For example, with some changes to bsnes's disassembler view, the following listing can be provided, as opposed to just a list of translated opcodes

![bsnes with custom disassembler changes](https://i.imgur.com/A3BKkcy.png)

[...]

PR #202, opened 2018-10-12
Add address-to-line mapping and source file list to Symbols file

CypherSignal:
* This PR resolves https://github.com/vhelin/wla-dx/issues/191

This PR adds extra information to the Symbols file output by wlalink to provide, most notably, detailed information relating addresses in the ROM file to specific source files and lines.

As a proof of concept, I've integrated support of this in a branch of bsnes-plus allowing for richer disassembly in the emulator, to better give the context of the source files for a ROM:

bsnes with custom disassembler changes

The expanded symbol information was also used this to flesh out support to implement VS Code's Debug Adapter Protocol for bsnes-plus, allowing for the user to interactively set breakpoints and indicate the location of the current PC (and other addresses in the current callstack) in the source file:

VSCode displaying breakpoints, current PC, and callstack

And aforementioned interactivity also demonstrated [here](https://www.youtube.com/watch?v=hkuMV-1LLmI)

Running down the changes:

* The symbol output is based on the existing List data generated during object file assembly, and so should not be a significant extra burden to support.
* During assembly, source-file checksums are calculated and included in the list data, to then be included in the output Symbol file. This allows for the emulator/vm to properly identify if source files match what they were when assembly occurred.
* "Base" bank offsets are now included in the list data, as this missing data was warping some output List file and Symbol file information.
* crc32 calculation was moved out to a separate file from hashmap and given a more public interface, as it is now utilized by code other than the hashmap itself.
* Inclusion of the addr-to-line data is _optional_ through an additional cli arg, "-A" - only acknowledged when generating WLA symbols, of course. This is such a large addition and change to the symbol file output, some emulators may be a bit naive in how they were loading the symbol files, and error out on the new data, so the new data is excluded by default.
* A small doc was written up describing what the output of the Symbols file is, and how it should be interpreted by consumers.


vhelin:
Nice, thank you! 👍

vhelin:
My GCC 7.3.0 gave some warnings about .8x in strings, in wlalink/write.c so I (hopefully) fixed those. I also updated CHANGELOG and FILE_FORMATS.

CypherSignal:
Oh, I missed that FILE_FORMATS was a thing, and didn't realize that the intermediate files had any versioning. Thank you for addressing those.

PR #203, opened 2018-10-12
pass_4.c had some CR character that git complains about

Neui:
It seems fdd7e37ae0906c08ed82edd90b7e0437b36629c9 introduced an CR in pass_4.c, which git complains about (since it wants to convert from CRLF → LF), which is annoying since when rebasing git complains about "modified changes".


vhelin:
Thanks!

PR #204, opened 2018-10-12
Fix stuff PR #202 broke

Neui:
Related to #202.

* `wlalink/write.c` tried to include `..\crc32.h`, this doesn't work on linux, since it uses `/`. (Windows should also allow `/`, I think)
* `doc/symbols.rst` subheaders doesn't had the subsections correctly aligned, so sphinx would complain and it wouldn't compile.

Not directly a thing the PR broke, but I put the information for emulator authors into a subsection rather than a section, since that info is about the symbols.

As a bonus, add the newly `doc/symbols.rst` to the main page, after the architectural overview, since I think it fits in there.

### DIFF
--- a/doc/symbols.rst
+++ b/doc/symbols.rst
@@ -32,35 +32,35 @@ This is a list of all Labels to sections of the ROM, such as subroutine location
 - Format specifier: ``%2x:%4x %s``
 
 [definitions]
---------
+-------------
 This is a list of various definitions provided in code - or automatically during WLA's processing - and values associated with them. Most prominently, WLA outputs the size of each section of the ROM. Each line lists an integer value in hexadecimal, and a string (name) associated with that value.
 
 - Regex match: ``[0-9a-fA-F]{8} .*``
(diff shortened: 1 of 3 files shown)


vhelin:
Nice, thank you! :D

vhelin:
The / fix works for me, although \ worked also. CygWin (latest), Windows 10 here...

Neui:
[Cygwin supports both `\` and `/` as directory seperator.](https://cygwin.com/cygwin-ug-net/using.html#pathnames-intro), which is probably why it accepted `\`. It seems the Windows API also accepts `/` (couldn't find it in the official documentation, only in other sites), but not (all) Windows command line tools.

cr1901:
> It seems the Windows API also accepts / (couldn't find it in the official documentation, only in other sites), but not (all) Windows command line tools.
This about hits the nail on the head, but internally Windows API functions can tolerate `/`, even if some command line tools choke.

PR #209, opened 2019-01-02
[spc700] Fix incorrect assembly

forarslys:
This will fixes the followings

- typos:
1.  **RET1** -> **RETI**
2. **AND1** mem.bit, C -> **MOV1** mem.bit, C
- incorrect 8-bit operand type. (most of 8-bit relative -> 8-bit reference)
- incorrect assembly with bit operations such as and1, or1, and etc.; which specify _13-bit address_ and which bit to be affected.

The last should be the most important fix. For 13-bit reference that is to be resolved in linking process, I newly adopted a letter 'n' and 'N', which seemed unused so far, for internal coding during the pass 3 and pass 4.
I'd appreciate reviews, since these commits involve the core of the program and I may be doing something wrong with lack of understandings.


vhelin:
Thank you! A great bundle of fixes! I don't know which old document I used back in the day to implement the old buggy code, but Google now gives me modern looking SPC-700 docs where the fixes you did are explained...

PR #210, opened 2019-01-03
Fix SPC-700 bit addressing instructions (MOV1, AND1, etc.)

rbolsius:
WLA 9.7 has a bug with SPC-700 instructions that use 13-bit byte addresses and 3-bit bit positions.  This includes MOV1, AND1, OR1, EOR1, and NOT1.  The operands of all these instructions are output incorrectly.

In the SPC-700 instruction set, bit addressing instructions such as AND1 have a 16-bit operand where the lower 13 bits are used for the byte address and the upper 3 bits are used to indicate the bit position within the addressed byte.  Since the address just occupies the lower 13 bits, only bytes from $0 to $1FFF can be accessed by these instructions.

Example 1:

```
AND1 C, $C8.1    ; opcode: 4A, address: $C8, bit position: 1
                 ; This will access the second bit in byte $C8, AND it with
                 ; the C flag, and write the result back to the second bit
                 ; in $C8.
```

Operand: 0xC8 | (1 << 13) = 0xC8 | 0x2000 = 0x20C8

Expected machine code:

    4A C8 20

WLA currently incorrectly outputs the bit position as the first octet and the byte address as the second octet.  Because of this, the byte address and bit positions are mixed up.

Incorrect output from WLA 9.7:

    4A 01 C8

Also, WLA 9.7 does not allow addresses above $FF to be used.

Example 2:

```
AND1 C, $200.3   ; opcode: 4A, address: $200, bit position: 3
                 ; This will access the fourth bit in byte $200, AND it with
                 ; the C flag, and write the result back to the fourth bit
                 ; in $200.
```

Operand: 0x200 | (3 << 13) = 0x200 | 0x6000 = 0x6200 

Expected machine code:

    4A 00 62

This example can't be assembled with WLA 9.7 since the address is larger than $FF.

This pull request solves these issues by implementing a new opcode type (0xF) for these bit addressing instructions.  In addition, the instruction MOV1 x.x,C (opcode 0xCA) was mistakenly entered in the SPC-700 opcode table as AND1 x.x,C.  This has also been fixed.


vhelin:
Hi!

There's a couple of days old pull request that I'm currently checking out:

  https://github.com/vhelin/wla-dx/pull/209

I haven't fully read through these pull requests, but are they doing the same thing?

vhelin:
I'll try to check both of these pull requests out completely tomorrow...

rbolsius:
Yes, it does look like #209 is addressing the same issues.  What are the odds that there would be two PRs for this in two days?  As far as I can tell, this issue has been there for a while.

Anyway, it looks like the PR from forarslys is more complete than mine.  That's for taking a look.

vhelin:
Ok! I approved pull request #209 as it looked good... If you find anything
to fix there, please let us know! Thanks!

vhelin:
And yeah, my old buggy code has been there for years, and now two
separate/independent pull request fixes to these matters inside two days -
mind blown. :D

PR #217, opened 2019-02-18
Small improvements

Neui:
Some small improvements I've done on the code while waiting to compile.

Also fixes the DOS DJGPP-GCC build since when printing the help, `tmp_name` is never set and the `atexit` handler then tries to use it, uninitialized and thus crashes.

### DIFF
--- a/defines.h
+++ b/defines.h
@@ -96,22 +96,6 @@
 
 #define OP_SIZE_MAX 16
 
-#ifdef AMIGA
-struct optcode {
-  char *op;
-  int  hex;
-  short int type;
-  short int value;
-};
-#else
-struct optcode {
-  char *op;
-  int  hex;
-  int  type;
-  int  value;
-};
-#endif
-
 #endif
 
 /**************************************************************/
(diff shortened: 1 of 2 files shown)


vhelin:
Thanks! I always thought that in ANSI C free(NULL); would crash, but now that I read the free()'s documents I see that's not the case...

cr1901:
>I always thought that in ANSI C free(NULL); would crash
@vhelin `free(NULL)` is _supposed_ to be a noop, even in ANSI C. AIUI, historically compilers didn't implement this correctly and indeed _would_ crash. Perhaps you have memories of this :D?

vhelin:
On Tue, Feb 19, 2019 at 3:42 AM William D. Jones <notifications@github.com>
wrote:

> I always thought that in ANSI C free(NULL); would crash
> @vhelin <https://github.com/vhelin> free(NULL) is *supposed* to be a
> noop, even in ANSI C. AIUI, historically compilers didn't implement this
> correctly and indeed *would* crash. Perhaps you have memories of this :D?
>
>
That could very well be the case. :)

PR #227, opened 2019-03-09
Normalize pass1.c

Neui:
Looks like on that line the ended with an CRLF and wasn't converted to LF (normalized).

BTW you probably want to change your `core.autocrlf` setting to either `true` (for having CRLF in the working directory) or `input` (for having LF in the working directory), but NOT `false`, since that doesn't normalizes it and git is going to complain on non-Windows systems and on Windows with it set to `true` or `input`.


vhelin:
Thanks!

PR #243, opened 2019-06-21
Unions 2: Electric Boogaloo (+ new DSTRUCT syntax)

Drenn1:
This is a redo of #174; a lot of stuff had been moved around, making it necessary to start basically from scratch.

Somewhat analagous to a C union, this allows you to reuse memory in an enum, ramsection, or struct. From the documentation:

```
    .ENUM $C000
        .UNION
            pos_lowbyte:  db
            pos_highbyte: db
            extra_word:   dw
        .NEXTU
            pos:          dw
        .ENDU
        after: db
    .ENDE
```

This example is equivalent to:

```
    .DEFINE pos_lowbyte  $c000
    .DEFINE pos_highbyte $c001
    .DEFINE extra_word   $c002
    .DEFINE pos          $c000
    .DEFINE after        $c004
```

This involved refactoring enums, ramsections, and structs especially. Enums and ramsections now make temporary structs, so all 3 of those things share the same code.

You can also add a name as an argument to ".UNION" or ".NEXTU" to put it into something similar to a namespace.

One somewhat important note, is that the behaviour of instanceof with DESC enums has changed. Specifically, if you had:

```
.ENUM $d000 DESC
    thing: instanceof MyStruct 3
.ENDE
```

Then, assuming "MyStruct" is 2 bytes long, before it would generate labels like this:

```
00:cffe thing
00:cffe thing.1
00:cffc thing.2
00:cffa thing.3
```

But now, it does this:
```
00:cffa thing
00:cffa thing.1
00:cffc thing.2
00:cffe thing.3
```

I can change it back if you want; but I think, at least, it makes more sense for "thing" to be $cffa rather than $cffe, since it should point to the beginning of all subsequent instances.

---

In addition, last time, I ran into the problem of .DSTRUCT not working with unions. I've made a completely new syntax for it as a result. The old syntax is still there as well, so as not to break compatibility. There's more information in the documentation, but, here's how it looks:

```
    .STRUCT water
        name   ds 8
        age    db
        weight dw
    .ENDST

    .DSTRUCT waterdrop INSTANCEOF water VALUES
        name:   .db "tingle"
        age:    .db 40
        weight: .dw 120
    .ENDST
```

.DSTRUCT actually calls the "parse_directive" function to handle data definitions, so it mostly behaves like a normal section. I felt that this was necessary, otherwise there was no flexible way to define large data sections (ie. if you had a dsb $100). Though of course, some directives, such as .ORG, are disabled here.

The field order does not matter with this syntax, and as a result, it works with unions. Anything that's omitted is filled with .EMPTYFILL. I've also added the option to use .DSTRUCT without a name (so, no labels are generated).

This should be ready to merge, but I can't guarantee it's bug-free. It's a fairly significant overhaul in certain places. If you're releasing v9.8 soon, you might want to do so before merging this, so the kinks can be worked out before v9.9. But it does compile my rather large pet project correctly, and the tests appear to be working, mostly. 65816/operand_hint_test is failing, but it's not my fault; an error check I added is being tripped. The same problem occurs on the master branch if that error check is added there.


vhelin:
> This should be ready to merge, but I can't guarantee it's bug-free. It's a fairly significant overhaul in 
> certain places. If you're releasing v9.8 soon, you might want to do so before merging this, so the kinks
> can be worked out before v9.9.

Releasing v9.8 before merging this sounds like a good plan, like you suggested.

vhelin:
Thanks! This is a big one. Good work. :)

PR #246, opened 2019-07-21
Improve support/docs for some (rare) GB header values

Gekkio:
* 64 Mbit ROM: supported by MBC5. In official games used by Densha De Go 2
* 512 kbit RAM: supported by MBC30 (variant of MBC3). In official games used by Japanese Pokemon Crystal
* (docs only) MBC6: In official games used by Net De Get 100
* (docs only) MBC7: In official games used by Command Master, and the two versions of Kirby Tilt 'n' Tumble
* Pocket Camera should have value $FC


vhelin:
Thanks!

PR #252, opened 2019-07-30
Allow string defines that have numbers as a prefix. (e.g. -D TEST=123test)

crzysdrs:
Normally when trying to define a numeric prefixed string you would get an invalid value error. Now, it will allow anything that isn't fully a decimal or hex value to be considered a string. 
Motivating Example: -D TEST=123test


vhelin:
If we allow 123test to be string from the command line for WLA, WLALINK's [definitions] should allow that as well, and the source code parser inside WLA... :)

crzysdrs:
It looks like the source code parser will accept definitions as strings for something like:
.define TEST "123test"
but WLALINK '[definitions]' is exclusively numeric (with no apparent storage/handling within a label for a string-like value).

I would be fine with instead allowing -D TEST="123Test" or some other quoting mechanism to indicate that the value is explicitly a string to bring it more in line with the source definitions. Updating WLALINK looks to be a more complicated endeavor.

vhelin:
> It looks like the source code parser will accept definitions as strings for something like:
> .define TEST "123test"
> but WLALINK '[definitions]' is exclusively numeric (with no apparent storage/handling within a label for a string-like value).

Ouch, didn't remember that. I guess there is no need for string definitions in WLALINK as they cannot be used in the linking time, .DBs etc. need their data in the assembler...

> I would be fine with instead allowing -D TEST="123Test" or some other quoting mechanism to indicate that the value is explicitly a string to bring it more in line with the source definitions. Updating WLALINK looks to be a more complicated endeavor.

-D TEST="123Test" sounds really good and is enough.

crzysdrs:
This modified version allows things on the command line like:
-D 'TEST="123test"'
-D 'TEST="123\"test"'
-D TEST=\"123test\"

It's worth noting that the single quotes are required to prevent the shell from removing the unescaped double quotes examples.

Double quotes that are escaped in the final string are handled in the same way that the normal string processor works, \" becomes ".

vhelin:
Thanks!

PR #184, opened 2018-06-27
Improved building via CMake

Neui:
This does somewhat a lot of stuff, mostly taken from #164:

* Generally cleaner and consistent `CMakeLists.txt`s
* Now using a list of architectures(/CPUs) rather than "hardcoding" the things, so now a loop can be used.
* Cross-compiling support, including basic support for "crosscompiling-emulators" (e. g. wine)
* CPack support
* Opcode generator now accepts an optional argument for the output file
* Opcode generator has been moved to the root directory because it is just one file and makes possible to use the output as a source file. (You cant specify generated source files from other directories in CMake for some reason)
  * The `create_tables.bat` also has been removed as a side effect. @cr1901 is probably going to complain about this.
* User code now explicitly refers the generated opcode table via a `extern` variable. I am not so sure if this is going to work on amiga (because of the `_far` thing).
* Define `NDEBUG` now also controls the debugging defines for wla and wlalink since `NDEBUG` is basically in the C standard (for `assert`)

Please review (and test) before merging. Complains allowed.

PR #254, opened 2019-08-17
Git: Make makefiles ending in .gb be text

Neui:
Because of the later `.gb binary`, git thought `smake.gb` and similar were binary files instead of text files. This fixes it for makefiles.

I've noticed this when rebasing #184 and it wouldn't merge the "binary files".

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -44,6 +44,7 @@ historical/wlad_new/makefile.win32* text eol=crlf
 *.exe   binary
 *.bin   binary
 *.gb    binary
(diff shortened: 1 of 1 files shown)


cr1901:
I didn't even know you could do this, thanks! Waiting for @vhelin to merge, looks good to me.

Neui:
I force-pushed to include the path, just in case an `.gb` file with "make" is accidentally commited.

vhelin:
Thanks!

PR #255, opened 2019-09-08
Make test failures obvious

strager:
When running wla-dx' tests, I was surprised that the run_unit_tests.sh script succeeded despite some errors being reported. run_unit_tests.sh basically never exits with a failure. This makes it difficult to see if a test actually failed; you need to sift through hundreds of lines of output for any interesting error messages.

Make run_unit_tests.sh fail as soon as a failure is encountered. This puts the error message at the bottom, and makes run_unit_tests.sh return a non-zero exit code if any test fails.

---

When running run_unit_tests.sh, you will only notice if wla produces incorrect binaries if you look carefully at the script's (very verbose) output. Make finding test failures easier by making byte_tester propagate a failure exit code up to run_unit_tests.sh.

### DIFF
--- a/byte_tester/main.c
+++ b/byte_tester/main.c
@@ -9,7 +9,7 @@ int main(int argc, char *argv[]) {
 
   char tmp[256], test_id[256], tag_id[256];
   unsigned char *binary_file, bytes[256];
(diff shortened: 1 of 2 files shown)


vhelin:
Thanks!

PR #256, opened 2019-09-08
Fix checksum for 24 Mbit SNES ROMs

strager:
For 24 Mbit SNES/65816 ROMs like EarthBound and Super Metroid, the .computesneschecksum directive produces the wrong checksum. The SNES checksum algorithm implements mirroring incorrectly; mirroring should round to the next power of two, but .computesneschecksum instead rounds to the nearest 4 Mbit boundary.

Fix the implementation of .computesneschecksum for LOROM, HIROM, and EXLOROM variants.

I manually verified that this change produces correct checksums for a variety of licensed SNES ROMs, including LOROM, HIROM, and SA-1 ROM variants. I did not test any EXLOROM or EXHIROM ROMs.

### DIFF
--- a/examples/65816/checksum_12mbit_lorom/linkfile
+++ b/examples/65816/checksum_12mbit_lorom/linkfile
@@ -1,0 +1,2 @@
+[objects]
+main.o
--- a/examples/65816/checksum_12mbit_lorom/main.s
+++ b/examples/65816/checksum_12mbit_lorom/main.s
@@ -1,0 +1,35 @@
+.lorom
+.memorymap
+	slotsize $8000
+	defaultslot 0
+	slot 0 $8000
+.endme
+
+.rombanksize $8000
+.rombanks $30
+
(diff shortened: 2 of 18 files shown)


vhelin:
Looks good otherwise, but are the new tests missing Makefiles? .gitignore is set to ignore makefiles, so every time I need to commit makefiles I'll have to edit it...

Neui:
> Looks good otherwise, but are the new tests missing Makefiles? .gitignore is set to ignore makefiles, so every time I need to commit makefiles I'll have to edit it...

I did a PR to "fix" this for the example directory. Note that only non-tracked ones needs to be added forcibly (git add -f). Already added makefiles (like existing ones) are already tracked and thus not ignored.

strager:
> Looks good otherwise, but are the new tests missing Makefiles?

Yes! Thanks for noticing. (And thanks @Neui for #257.) I'll correct this.

strager:
(linkfile is also missing (because of .gitignore), and #257 doesn't fix that.)

vhelin:
Thanks!

PR #258, opened 2019-09-11
Fix corrupt NO$SNS symbol file for 24-bit symbols

strager:
When a non-label symbol has a 24-bit value, NO$SNS fails to load the .sym file generated by wla-dx, reporting the following error:

> Symbolic Info File Corrupt,
> Line   247

This happens because the .sym file contains bad data:

    000088b8 _sizeof_a
    000056fc _sizeof_b
    00004310d _sizeof_c
    000016eba _sizeof_d
    00004000 _sizeof_e
    00001600 _sizeof_f

Fix wla-dx' formatting of the .sym file so NO$SNS can load them.

### DIFF
--- a/examples/65816/nocashsns_symbol_file/check_symbols.sh
+++ b/examples/65816/nocashsns_symbol_file/check_symbols.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+set -e
+set -u
+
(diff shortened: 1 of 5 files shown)


vhelin:
Thanks!

Issue #110, opened 2016-05-31
Address Relative to Label Issues


TheMessengerDevelopments:
EDIT: I've found this problem in version 9.7 as well, which I just cloned and compiled today.
I've recently started playing with wla-65816 v9.5, and I just noticed some unexpected behavior with the `PER` operator, specifically when adding and subtracting from it's operand label:

```
PER break-1
BRA sampleRoutine
break:
...
sampleRoutine:
...
RTS
```

This simple bit of code was intended to push the address of the byte preceding `break`, then jump to sampleRoutine, then jump back the the byte before break to continue on with whatever is in the `...` (the program counter gets incremented automatically after the `RTS`).  
I expected the assembler to evaluate `PER break-1` to `PER $0002` as `BRA` is 2 bytes in size.  However, when I looked at the assemblers output in a debugger (I'm using Geiger's snes9x debugger), I found that it actually output was `PER $0003`.  I ran some more tests, leading to these results:
Assembly -> Output
`PER break-3` -> `PER $0001`
`PER break-2` -> `PER $0002` (what I wanted)
`PER break-1` -> `PER $0003` (as above)
`PER break` -> `PER $0003`
`PER break+1` -> `PER $0005`
I can't be the only one who though that looked weird.  It seem to me that the relative address in question is offset by one, but only after arithmetic.  This only seemed to happen with `PER`, as I tried the same experiment with `PEA`, and got the expected behavior (probably because of relative vs absolute addresses). 
Anyway, does anyone know what's going on here? Is this supposed to happen, or is it some bug that by some twist of fate was never noticed?  And does anyone know any way to make it work in a way that doesn't seem so weird?  Your responses are much appreciated.  

PS. Thank you developers for making this awesome compiler, as it has provided my many hours of fun.  Also, I hope you enjoyed your trip Mr. Helin.

nicklausw:
I'll look into this in a sec, but note: do not use v9.5, as that is not the most recent version. Use the latest version from here (I can probably give binaries if needed).

TheMessengerDevelopments:
I see. I didn't realize this was still being so actively developed and getting new versions.  Thanks!

TheMessengerDevelopments:
Also a binary would be nice.  Or even just some help compiling it.  I can't figure out how to compile c++ to save my life.

nicklausw:
[...]

PR #259, opened 2019-09-16
Fix pending calculations of 16-bit relative offsets (#110)

mrehkopf:
65816 and 6809 support relative addressing with 16-bit displacements (PER / BRL; LBxx) which is currently broken for pending calculations (e.g. a popular case with 65816 is `PER label-1`) because a relative pending calculation always assumes an operand size of 8 bits. Consequently the calculated relative address will be off by one byte.

When calculating the relative operands take into account the size of the operand to obtain the correct PC-relative offset.
The operand size is already passed by the assembler as the stack type, so use it to adjust the result of relative calculations.

### DIFF
--- a/wlalink/write.c
+++ b/wlalink/write.c
@@ -2025,7 +2025,7 @@ int parse_stack(struct stack *sta) {
   struct section *s;
   struct label *l, lt;
   double k;
-  int g;
+  int g, ed;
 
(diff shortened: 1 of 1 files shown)


vhelin:
Amazing and thorough work! 👍 Thanks!

PR #266, opened 2019-11-27
Better README.md

Neui:
Had it laying around, now a bit changed and added information about the install prefix.


vhelin:
Thanks!

Issue #273, opened 2019-12-20
Usage of sprintf across codebase


AntonioND:
Today I was building WLA-DX in my machine and got the following warning:

```
wla-dx/wlalink/main.c: In function ‘parse_and_set_libdir’:
wla-dx/wlalink/main.c:861:27: warning: ‘sprintf’ may write a terminating nul past the end of the destination [-Wformat-overflow=]
  861 |   sprintf(ext_libdir, "%s/", n);
      |                           ^
wla-dx/wlalink/main.c:861:3: note: ‘sprintf’ output between 2 and 257 bytes into a destination of size 256
  861 |   sprintf(ext_libdir, "%s/", n);
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The problem is that the compiler realizes that the max possible output of ``sprintf()`` is too big for the buffer.

In this case, it is easy to fix the problem by increasing the size of ``ext_libdir``, for example, and I'm more than happy to submit a PR. However, I checked with grep and it looks like all the codebase is using ``sprintf()`` rather than ``snprintf()``. I guess it's not a huge deal because this tool isn't really security-critical, and the compiler doesn't complain about other instances of ``sprintf()`` so in most cases it seems to be fine (but the checks don't work when ``%s`` refers to a ``char*``, it only works with arrays, so there are surely more problems like this.

I think that in general all the calls should be checked, and anything that comes as input from the user should use ``snprintf()`` to make sure that there are no overflows. What do you think?

In case you're wondering, I'm using ``gcc (Debian 9.2.1-21) 9.2.1 20191130``.

Neui:
This is known, but there is no `snprintf` in C89, that was introduced in C99, and the author wants to stay on C89 for compatibility. Increasing buffer sizes would help, but not solve the problem.

I have thought of maybe using some wrapper function and use `snprintf` when available (detect by platform/build system), but I haven't thought further.

vhelin:
We really need to continue using C89 as I want to keep e.g. Amiga support. :) But perhaps someone could write a similar function to snprintf()? I'm sure the use cases of sprintf() in WLA DX are all very similar and quite simple so writing a function that does the same shouldn't be a huge issue in our case...

cr1901:
>I have thought of maybe using some wrapper function and use snprintf when available (detect by platform/build system), but I haven't thought further.

This would also be my suggestion in the future, though for now I would recommend the following:

[...]

PR #286, opened 2020-01-28
Make buffer large enough for sprintf's output

SolraBizna:
Compiling current master with GCC gives the following warning:

```
.../wlalink/main.c: In function ‘parse_and_set_libdir’:
.../wlalink/main.c:897:27: warning: ‘sprintf’ may write a terminating nul past the end of
the destination [-Wformat-overflow=]
   sprintf(ext_libdir, "%s/", n);
                           ^
.../wlalink/main.c:897:3: note: ‘sprintf’ output between 2 and 257 bytes into a
destination of size 256
   sprintf(ext_libdir, "%s/", n);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is trivial to fix by increasing the size of `ext_libdir` by one byte. (Really, `snprintf` should be used, but I assume there's some historical reason it isn't being used.)

### DIFF
--- a/wlalink/main.c
+++ b/wlalink/main.c
@@ -60,7 +60,7 @@ int num_sorted_anonymous_labels = 0;
 extern struct section_fix *sec_fix_first, *sec_fix_tmp;
 extern char mem_insert_action[MAX_NAME_LENGTH*3 + 1024];
 extern int emptyfill;
-char ext_libdir[MAX_NAME_LENGTH + 1];
+char ext_libdir[MAX_NAME_LENGTH + 2];
 
 
 #ifdef WLALINK_DEBUG


Neui:
> (Really, `snprintf` should be used, but I assume there's some historical reason it isn't being used.)

See #273

> [...] there is no `snprintf` in C89, that was introduced in C99, and the author wants to stay on C89 for compatibility. Increasing buffer sizes would help, but not solve the problem.
> [... Project author:] We really need to continue using C89 as I want to keep e.g. Amiga support.

SolraBizna:
I expected it to be something like that. I'm personally fortunate enough to be able to count on C11 and C++14 across my own projects. (Very weird to go back and forth between C++14 and 6502 assembly on the same project.)

vhelin:
Anybody wants to code (a limited version, we don't need all its features) snprintf() for this C89 project? :) I'm busy implementing new features and fixing bugs...

Issue #293, opened 2020-02-16
Errors with getting label bank number


Drenn1:
I really have no idea what's going on here, so I'll just let the code speak for itself.

```
.include "cgb_hardware.i"

.MEMORYMAP
SLOTSIZE $4000
DEFAULTSLOT 1
SLOT 0 $0000
SLOT 1 $4000
SLOTSIZE $2000
SLOT 2 $c000
.ENDME

.ROMBANKSIZE $4000
.ROMBANKS 2

.emptyfill $ff

.ORGA $150

myLabel:
	.db $01 :myLabel ; Silent failure: only $01 is defined
	.db $01, :myLabel ; This works

	.db :$01 | :myLabel ; Gives the error below:
	; _GET_BANK_OF_ADDRESS: SLOT -1 < 0! Internal error. Please submit a bug report.
	; main.o: main.s:29: COMPUTE_STACK: Could not get the bank number for ROM address 1/$1 (out of bounds).

```

vhelin:
.db $01 :myLabel ; Silent failure: only $01 is defined

Due to a recent change, ':' is now a operation for the internal stack calculator, thus this transforms into a stack calculation of $01, OP_GET_BANK_NUMBER, myLabel -> as we parse through the stack there is nothing that would join the first $1 and OP_GET_BANK_NUMBER(myLabel), $1 is just what remains...

Drenn1:
Remember that parser rewrite that only supported commas for macro delimiters? The guy might have had a point =P

vhelin:
Duh... :D 

The error related to this one:

.db :$01 | :myLabel

is because a plain $01 has no SLOT associated to it. If it was :(myLabel+1) we'd know it's ROM bank... Without the SLOT association, perhaps I can just use .ROMBANKSIZE to calculate the BANK for this address?

Drenn1:
Interesting. I, uh... didn't mean to write ":$01", I meant to write "$01". Whether you consider that a bug or not is up to you...

This is the REAL bug:

```
	.db $01 | :myLabel ; Gives the error below:
; main.o: main.s:23: COMPUTE_STACK: OR is missing an operand.
```

vhelin:
That's a parser feature. 

.db $01 | (:myLabel)

should work.

vhelin:
Ok, I might have managed to fix the first bug in this report... Can you verify?

Drenn1:
[...]

PR #294, opened 2020-02-16
Better handling of unary operators

Drenn1:
Possible fix for #293. The jist of this is that you should be able to do ".db $01 | :label" without having to wrap ":label" in brackets. Same with ">" and "<" operators.

I don't _entirely_ know what I did, so I guess take this with a massive grain of salt? I just sort of copied the way multiplication appears to handle precedence. But all of the tests pass, and my massive project compiles, so it doesn't seem too broken?

Review thoroughly.

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -646,7 +646,8 @@ int stack_calculate(char *in, int *value) {
 
   /* fix the sign in every operand */
   for (b = 1, k = 0; k < q; k++) {
-    if ((q - k) != 1 && si[k].type == STACK_ITEM_TYPE_OPERATOR && si[k + 1].type == STACK_ITEM_TYPE_OPERATOR && si[k + 1].value != SI_OP_BANK) {
+    if ((q - k) != 1 && si[k].type == STACK_ITEM_TYPE_OPERATOR && si[k + 1].type == STACK_ITEM_TYPE_OPERATOR && si[k + 1].value != SI_OP_BANK
+        && si[k + 1].value != SI_OP_HIGH_BYTE && si[k + 1].value != SI_OP_LOW_BYTE) {
       if (si[k].value != SI_OP_LEFT && si[k].value != SI_OP_RIGHT && si[k + 1].value != SI_OP_LEFT && si[k + 1].value != SI_OP_RIGHT) {
 	print_error("Error in computation syntax.\n", ERROR_STC);
 	return FAILED;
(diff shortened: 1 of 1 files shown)


vhelin:
This is awesome, but I cannot accept this PR today as I don't really understand what it does either. :D This is what I was talking about previously, I have forgotten how the stack calculator works in WLA DX, and need to relearn how it works... I'll try to spend time tomorrow understanding this. But if this works 100%, good work!

vhelin:
This is how the SI_OP_MULTIPLY behaves here:

```
        else if (si[k].value == SI_OP_MULTIPLY) {
          b--;
          while (b != -1 && op[b] != SI_OP_LEFT && op[b] != SI_OP_PLUS && op[b] != SI_OP_MINUS) {
            ta[d].type = STACK_ITEM_TYPE_OPERATOR;
            ta[d].value = op[b];
            b--;
            d++;
          }
          b++;
          op[b] = SI_OP_MULTIPLY;
          b++;
        }
```

But this is how SI_OP_BANK works after this PR:

```

else if (si[k].value == SI_OP_BANK) {
	  b--;
	  op[b] = SI_OP_BANK; /* Unary operator */
}

```

Ah, I need to study this more..

Drenn1:
I initially thought, what if I just added "&& op[b] != SI_OP_OR" to the while loop. The idea being that since multiplication checks "SI_OP_ADD", it's probably overriding that for precedence. And that worked.

But unary operators should override _everything_ before them, right? So we're going to end up checking everything in the while loop - which just means the while loop can be removed. That was my thought process anyway.

Drenn1:
Although I'm not sure where you got that second code snipped from, it should look like this:

```
	else if (si[k].value == SI_OP_HIGH_BYTE) {
	  op[b] = SI_OP_HIGH_BYTE; /* Unary operator */
	  b++;
	}
```

vhelin:
> Although I'm not sure where you got that second code snippet from, it should look like this:
> 
> ```
> 	else if (si[k].value == SI_OP_BANK) {
> 	  op[b] = SI_OP_BANK; /* Unary operator */
> 	  b++;
> 	}
> ```

Oops, my copy paste error. That's what I meant... Still, codewise very different from what SI_OP_MULTIPLY does. But after reading your explanation I can see what you did there...

Let's see, later next week I should be able to understand all this and make decisions... :)

vhelin:
Great work, I think your code works! :) After reading this

  https://cs.nyu.edu/courses/fall14/CSCI-GA.1133-001/HW/postfix-calculator.html

(the important part is "The Converter Class") I think your code does just that - gives '<', '>' and ':' priority over everything else.

PR #312, opened 2020-05-16
Accept "appendto" in linkfiles

Drenn1:
Turns out I may have a use case for "appendto" after all. Seems cleaner to do it in the linkfile though!

I noticed while implementing this that APPENDTO has issues when there's a chain of length 3 or more. They need to be defined from first to last for it to work properly. Not sure how best to fix it. I'll make a separate issue.


vhelin:
Thanks!

PR #317, opened 2020-05-27
Fix CMake build as a subdirectory

rdrpenguin04:
I am embedding wla-dx into the source tree of my build, and I figured out that CMake wasn't building. This is because some of the locations of generated files are coded to CMAKE_BINARY_DIR, when they should be CMAKE_CURRENT_BINARY_DIR.

This PR should fix the issue. It builds correctly on my machine, hoping Azure works as well.


cr1901:
Oh, so _that's_ the difference between the two. This LGTM. Will wait to see if Ville has any feedback.

vhelin:
I know very little of CMake, so if you @cr1901 think this is good then let's merge it!

About Azure, we shall see what happens. :)

vhelin:
Building seems to work in Azure, so no problem there.

Issue #329, opened 2020-08-07
wlalink: symbol file name generation incorrect


alasher:
I noticed something peculiar when generating the WLA symbol files, sometimes they would strangely be named only `.sym`, I was only able to find them after checking `git status`. I dove into the code and found the problem: the code that modifies the file extension to `.sym` finds the _first_ dot character in the string. But really, we want the _last_ dot. My issue was I was using the current-directory syntax for my output file, ie. `./program.gb`.

While looking through this code, I noticed another problem: the new `name` array we made is only length 256, so if the `outname`'s length happens to be greater than this, it would write outside the bounds of the array. Yikes.

I decided to fix both problems as they were pretty simple, I will submit a pull request referencing this issue.

vhelin:
Good work! Thanks for the patch!

PR #330, opened 2020-08-07
Fix wlalink issue selecting symbol filename (#329)

alasher:
I fixed the problems listed in the issue (#329) and retained the existing functionality of appending the extension to the end of the filename if no dot exists within the string. I also removed the seemingly unnecessary `p` variable. Please feel free to let me know if you have any concerns with this change!


vhelin:
Thanks!

PR #338, opened 2020-12-01
Explicit casts for MSVC warnings on implicit narrowing

maxim-zhao:
A stock load into Visual Studio 2019 as a CMake project gives a bunch of warnings on implicit narrowing assignments. This just adds explicit casts to all of them to get rid of the warnings.


vhelin:
Thanks!

Issue #263, opened 2019-10-23
(Proposal) Variable-width string mapping


maxim-zhao:
This is a sort of extension to the `.asciitable` functionality, where instead of mapping single characters to single bytes, we instead map variable numbers of characters to variable numbers of bytes. The translation would match the longest possible "key" at the current point in the input string, and then emit whatever the output data is for that.

There are a few motivations:

1. Variable length strings are a good substitute for UTF-8 support - ANSI C code can see "オ" in UTF-8 as three or so chars and we can then map that to a single byte translation.
2. It could be useful for placeholders, e.g. I can map ASCII to 0-96 and then have "control sequences" like `<item>` map to single-byte codes above that range - nicer than having to inject `.db`s in the middle of my `.asc`s
3. We might actually want variable-length translations too
4. Sometimes I want 16-bit outputs for my `.asciitable` and the filter-macro version is really ugly

I can see that it might be painful to implement the "range" feature of `.asciitable` here because byte sequences from UTF-8 might well not come out as intended - although sequential Unicode codepoints will have mostly-sequential trailing bytes, there will be jumps every 64 code points. However the long-hand version is not the end of the world (people already make .tbl files this way).

...speaking of which, maybe this can be supported by reading a .tbl file? This tends to be in the form `<hex>=<text>`, where `<hex>` can be multiple bytes and `<text>` is some sort of text which may evaluate to multiple non-control-character bytes.

maxim-zhao:
Possible example syntax...
````
.dstringmap script ; named map because I may need more than one
; UTF-8 means the values are multi-byte
$00="ァ"
$01="ア"
$02="ィ"
$03="イ"
; ...
; Control codes
$ff01="<number>"
$ff02="<player>"
$ff03="\n" ; Doesn't matter if this is interpreted as one or two chars
.endd

.stringmap script "アイ<player>\n"
````
I'd suggest making it an error if you encounter any unmappable characters in the string.

It's a bit overwhelming to consider adding this myself... are there any docs around that might guide me towards things like how to add directives and the collection types WLA DX already has?

vhelin:
Sounds like an interesting feature!

I don't think we have any guides or documentation around about adding new directives to WLA DX. The starting point for a directive is in pass_1.c. Search for e.g., "directive_asctable_asciitable" in there for reference.

maxim-zhao:
[...]

PR #339, opened 2020-12-03
String mapping using TBL files, implements #263

maxim-zhao:
This implements the "variable-width string mapping" proposed in #263, see there for the rationale. It makes use of the somewhat standard TBL file format for defining bytes <-> string mapping, allowing use of the same file for script dumping and insertion. 

Note that I attempted to match the ANSI C requirements but don't have an enforcing compiler to check it. I also tried to match the code style but found it necessary to add some comments too 😄. There is a test (of sorts, I added it to an existing one) and documentation. Feedback is welcome.


maxim-zhao:
I didn't implement inline definitions of the table - it seems like it should be possible to share the code to work either way but I couldn't think how.

vhelin:
Good work! In fact the number of new code lines is much smaller than what I thought that implementing this feature would require. Well, I have no experience on working with UTF-8 etc., other than ASCII... :)

One question about the .tbl-file: Currently the format is

00=あ

but as the first number is a hex number, and hex numbers in WLA DX are of the form $00, should this be changed to

$00=あ

Also: should it support also decimal values? That'd make using it very clear without confusion... 

Bonus: add support for the other hex number type

00h=あ

? :)

maxim-zhao:
I'm sticking with the standard for TBL files which is undecorated hex. That way people can use the same file for multiple tools.

Issue #342, opened 2020-12-25
Bad memory write in macro_start_dxm


Drenn1:
My project has suddenly started getting segfaults. Investigation with valgrind gives me the following results on the latest commit 966f9c167e926be3a6da91d5eab13dee81f92031:

```
==552564== Command: wla-dx/binaries/wla-gb -o build/ages.o -DBUILD_VANILLA -D ROM_AGES ages.s
==552564==
==552564== Invalid write of size 8
==552564==    at 0x118390: macro_start_dxm (pass_1.c:353)
==552564==    by 0x122F06: directive_dbm_dwm_dlm (pass_1.c:2434)
==552564==    by 0x122F06: directive_dbm_dwm_dlm (pass_1.c:2398)
==552564==    by 0x1284AA: evaluate_token.part.0 (pass_1.c:1062)
==552564==    by 0x128A8A: evaluate_token (pass_1.c:1057)
==552564==    by 0x128A8A: pass_1 (pass_1.c:662)
==552564==    by 0x111594: main (main.c:186)
==552564==  Address 0x64b018c is 220 bytes inside a block of size 528 free'd
==552564==    at 0x483B9AB: free (vg_replace_malloc.c:538)
==552564==    by 0x11D3E3: undefine (pass_1.c:1309)
==552564==    by 0x11D3E3: directive_endm (pass_1.c:6903)
==552564==    by 0x1284AA: evaluate_token.part.0 (pass_1.c:1062)
==552564==    by 0x128A8A: evaluate_token (pass_1.c:1057)
==552564==    by 0x128A8A: pass_1 (pass_1.c:662)
==552564==    by 0x111594: main (main.c:186)
==552564==  Block was alloc'd at
==552564==    at 0x483CB65: calloc (vg_replace_malloc.c:760)
==552564==    by 0x1176BE: add_a_new_definition.part.0 (pass_1.c:1334)
==552564==    by 0x117929: add_a_new_definition (pass_1.c:1321)
==552564==    by 0x117929: redefine (pass_1.c:1282)
==552564==    by 0x117A4C: macro_start (pass_1.c:329)
==552564==    by 0x128F0D: pass_1 (pass_1.c:814)
==552564==    by 0x111594: main (main.c:186)
==552564==
==552564== Invalid write of size 4
==552564==    at 0x1183A1: macro_start_dxm (pass_1.c:360)
==552564==    by 0x122F06: directive_dbm_dwm_dlm (pass_1.c:2434)
[...]

PR #343, opened 2020-12-26
Potential fix for issue #342 (fix segfaults with .dbm/.dwm/.dlm)

Drenn1:
I don't entirely understand the code I modified, so don't pull unless you feel ok with it. I have a decent grasp on what the problem is, but I'm not 100% confident about the solution.

I don't know why the null checks were there in the first place. It seems to be like the code wrapped inside them is actually important initialization. But I could be wrong.

Also, regarding the "mrt->offset += 3" line; I have no idea what this does, or any of the other parts of that chain of if/else statements. It just looks to me like a case for the ".DBL" macro is missing. Again, don't pull this unless you're satisfied that this change makes sense.

I don't expect you to handle this today of course, enjoy your christmas =P


vhelin:
That mrt->offset is the current byte the macro is processing. Actually it should just be mrt->offset++ for .DBM, .DWM and .DLM, so there was a bug! :)

vhelin:
Also it makes sense to have macro_stack_grow() at the beginning of the function, like in macro_start_incbin()...

vhelin:
Hmm, I manually merged your pull request fixing the conflicts, and committed that, but it didn't close this pull request?

PR #344, opened 2021-01-03
Don't write defines to no$ style symbol files

Drenn1:
I fail to see why defines should be included in nocash-style symbol files. This results in lines like this:

```
00:0008 _sizeof_sound58Channel2
```

This is very misleading because it looks like a label, but it is not, and there is no way to distinguish the two.

The motivation behind this PR is that neither style of symbol file (WLA or nocash) works with bgb as of version 1.5.8. Or rather, they may work, but all "_sizeof" definitions pollute the list of labels to the point of being unusable. (In my case it flat-out refused to load the file at all, for some reason, but it seems to have something to do with this.)

While it may be BGB's fault that it no longer loads WLA-style symbol files properly, writing defines to nocash-style symbol files seems wrong, as they are indistinguishable from labels, making the file useless for debugging - so that's on WLA.

The result of this change is that only labels with status "LABEL_STATUS_LABEL" or "LABEL_STATUS_STACK" are written to nocash-style symbol files (LABEL_STATUS_DEFINE is ignored). I'm not sure what "LABEL_STATUS_STACK" means so I didn't change that.

### DIFF
--- a/doc/linking.rst
+++ b/doc/linking.rst
@@ -96,8 +96,8 @@ useful when you work under MSDOS (NO$GMB is a very good Game Boy emulator for
 MSDOS/Windows) as it contains information about the labels in your project.
 
 If flag ``S`` is used, WLALINK will create a WLA symbol file, that is much
-like NO$GMB symbol file, but shows also symbols and breakpoints, not just labels
-and definitions.
+like NO$GMB symbol file, but shows also symbols, defines, and breakpoints, not
+just labels.
 
 If flag ``d`` is used, WLALINK discards all unreferenced ``FREE``, ``SEMIFREE``,
 ``SEMISUBFREE``, ``SUPERFREE`` and ``RAM`` sections. This way you can link big
--- a/wlalink/write.c
+++ b/wlalink/write.c
@@ -1403,7 +1403,7 @@ int write_symbol_file(char *outname, unsigned char mode, unsigned char outputAdd
 
     l = labels_first;
     while (l != NULL) {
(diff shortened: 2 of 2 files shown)


vhelin:
LABEL_STATUS_STACK is a definition that contains a calculation (computation stack). It's good to filter those away as well in this case...

vhelin:
I don't know how to merge this pull request. The Linux CI is failing, it seems, for this one, perhaps it needs to be fixed first? But the error message doesn't really say anything...

???

Drenn1:
https://dev.azure.com/villehelin0486/villehelin/_build/results?buildId=238&view=logs&j=6b902995-b73d-5f5c-66fd-a7f66c857d2c&t=7cbfe919-8ac5-5a5d-dd4b-f8ecdf2baae9

One of the tests is failing. I guess the "nocashsns_symbol_file" test needs to be updated.

vhelin:
Good work, I didn't find the error that deep in the system. :) Fix is building...

vhelin:
Thanks!

Drenn1:
Most welcome, thank you for pulling!

vhelin:
No problemo, it was a good one, albeit small one. :) I'll try to check out
your other pull request early next week, I'm just totally busy @ work. I'm
100 meters deep in some javascript/python/erlang/go -mess that they put me
to hack, I'm totally out in another universe with this one... :D

Drenn1:
Are you going to filter out LABEL_STATUS_STACK as well? I didn't have time to do that before you merged it.

vhelin:
Good point. I think WLALINK has processed all those by the time we write the symbol files (and perhaps removed them?), but I'm not 100% sure. I'll filter them out just to be sure...

PR #345, opened 2021-01-04
Fixed bad memory access & bug with addr-to-line mapping

Drenn1:
The [addr-to-line] section in .sym files was entirely borked.


vhelin:
Thanks! I have to admit that listfile generation is a bit cryptic in WLA, but I think your fixes are correct!

PR #346, opened 2021-01-07
Fix VS Project Configs

jwrayth:
Only Release|Win32 configs were configured correctly. I have mirrored the appropriate changes to all other configurations and confirmed they now all build correctly.

Intermeddiary files are also added to the ignore file.


vhelin:
Thanks!

PR #349, opened 2021-01-08
Add support for multiple include directories via the -I flag

jwrayth:
This change allows the command line parsing to support multiple uses of the -I flag, allowing for multiple include directories to be added.

When searching for includes, the old pattern of priorities applies (External > incdir > current), however the externals are now processed in order of inclusion on the command line.

It is backwards compatible with existing uses of the -I flag.

I required this change locally as I have a both an external code library and an external asset library that is shared between multiple projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ ImportGenerators.cmake
 *.ilk
 *.ipdb
 *.iobj
+*.recipe
--- a/defines.h
+++ b/defines.h
@@ -541,6 +541,12 @@ struct section_def {
   struct section_def *next;
 };
 
+struct ext_include_collection {
+	int  count;
+	int  maxNameSizeBytes;
+	char** names;
+};
+
 struct incbin_file_data {
   struct incbin_file_data *next;
   char *data;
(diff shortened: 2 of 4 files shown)


vhelin:
The idea is great, I appreciate it, and will merge the pull request as soon as all problems are solved!

But
- The CI fails, can you fix those?
- You are using camel case for some new variables while the coding convention in this project calls for hello_world like variables... Please keep the style as it is.

After these are fixed I'll check out this pull request!

PS. I really hope we can integrate your pull request, the feature is nice!

vhelin:
Actually I can do the fixes myself... :)

jwrayth:
Ah thanks! I was going to look at it tonight, but busy fixing my car today.

jwrayth:
I don't like leaving a mess so I took a look at the errors shown. I wasn't seeing that locally as an error, but as a warning.

I'm on Windows and was building via Visual Studio, so there appears to be a configuration difference between that and the automated testing as whilst I did see that error locally, it was only classified as a warning. 

As there were many pre-existing warnings I didn't address it as it seemed this was not a priority element for this project.

I'll see about making Visual Studio and the tests more equal with each other in terms of strictness.

PR #352, opened 2021-01-10
Attempting to bring Visual Studio in line with other configs

jwrayth:
CMakeLists - Add all the header files to the WLA_SRCS so they are included in the Visual Studio projects. Set the W3 flag explicitly, but also set WX to force warnings as errors. Note that the "replace" method used for the O flag is not suitable here as W is not in the default flags.

byte_test/main.c - Build fix for warning treated as error on windows.

Outstanding Questions
- I'm not setup to test other build platforms/configs, so I am unsure if adding the headers will break other configs. It is very useful to see them in the VS Projects though
- Enabling W4 has nearly 1000 warnings, and at this time I'm assuming that is not a priority so I left disabled
- The CMake generated Visual Studio solutions do not match the layout of the repo submitted Visual Studio solution and project. To ensure this is kept up to date, it may be worth changing the hierarchy of that so it matches the cmake generated version.

If that last point is acceptable, I'll make that in a separate change after this. It'd make it much easier to keep the VS solution


vhelin:
> CMakeLists - Add all the header files to the WLA_SRCS so they are included in the Visual Studio projects. Set the W3 flag explicitly, but also set WX to force warnings as errors. Note that the "replace" method used for the O flag is not suitable here as W is not in the default flags.

Sounds good!
 
> byte_test/main.c - Build fix for warning treated as error on windows.
> 
> Outstanding Questions
> 
>     * I'm not setup to test other build platforms/configs, so I am unsure if adding the headers will break other configs. It is very useful to see them in the VS Projects though

I can test this on Linux and with Cygwin...

>     * Enabling W4 has nearly 1000 warnings, and at this time I'm assuming that is not a priority so I left disabled

Not a priority, but if you can see some potentially dangerous warnings among those ~1000 warnings, fixing them would be nice. :) Fixing all of them might be too much work versus what's to be gained...

>     * The CMake generated Visual Studio solutions do not match the layout of the repo submitted Visual Studio solution and project. To ensure this is kept up to date, it may be worth changing the hierarchy of that so it matches the cmake generated version.

I created the current Visual Studio solution manually as I didn't know at that time that you could create one automatically using CMake (I really don't know much about CMake). If solution generation with CMake is easy, then I suggest this:

- Add a note about how to generate Visual Studio solution with CMake to INSTALL
- Delete the old solution and stuff inside windows folder, and create a new solution there with CMake
- Check out if azure-pipelines-windows.yml needs updating

Or would something Visual Studio related (only in Visual Studio options?) in the solution be lost by doing this?

> If that last point is acceptable, I'll make that in a separate change after this. It'd make it much easier to keep the VS solution

Sounds good!

vhelin:
Thanks, at least Cygwin and Linux building works just fine with the modified CMakeLists.txt...

jwrayth:
> > ```
> > * Enabling W4 has nearly 1000 warnings, and at this time I'm assuming that is not a priority so I left disabled
> > ```
> 
> Not a priority, but if you can see some potentially dangerous warnings among those ~1000 warnings, fixing them would be nice. :) Fixing all of them might be too much work versus what's to be gained...
> 

I'll chip away at them in between other things :)

> > ```
> > * The CMake generated Visual Studio solutions do not match the layout of the repo submitted Visual Studio solution and project. To ensure this is kept up to date, it may be worth changing the hierarchy of that so it matches the cmake generated version.
> > ```
> 
> I created the current Visual Studio solution manually as I didn't know at that time that you could create one automatically using CMake (I really don't know much about CMake). If solution generation with CMake is easy, then I suggest this:
> 
> * Add a note about how to generate Visual Studio solution with CMake to INSTALL
> * Delete the old solution and stuff inside windows folder, and create a new solution there with CMake
> * Check out if azure-pipelines-windows.yml needs updating
> 
> Or would something Visual Studio related (some specific Visual Studio only options?) in the solution be lost by doing this?

The main difference between the CMake generated version and the Repo version is;

- Repo Version
  - Targets x86 and x64
  - Has Release and Debug configs only
  - Does not define MSDOS, so does not use DOS path delimiters and EOL characters
  - Targets Windows XP as lowest common platform
- CMake Version
  - Targets x64 Only (At least when generated on Windows 10 x64 with VS2019)
  - Has Debug, Release, MinSizeRel and RelWithDebug configs
  - Defines MSDOS, so uses DOS path delimiters and EOL and characters
  - Targets current version of Windows

The CMake can be updated to set the correct platform toolset to target Windows XP if that is still a hard requirement, and once that is done I assume a x86 build config could be generated also. I'm not sure if defining MSDOS is correct unless it is expected to run on DOS.

Is there a master list of platforms/configs that it is intended to run on? I'm happy to take some time working on the Windows side of things as that is my preferred development platform.

vhelin:
Yes, let's still keep supporting Windows XP! I remember getting an email a few months back about a user that still was using XP...

I don't think there is a master list of supported platforms, but here are the most important, I think:
- Windows XP
- Linux (as many distros as possible)
- macOS (as many versions as possible)
- AmigaOS 2.0
- MSDOS (don't know which one)...

... plus I'm developing WLA DX under Windows 10 and Cygwin, so that needs to work as well. :) 

If you want to look at MSDOS/AmigaOS build files for more details, they are in the historical folder... After every release I build AmigaOS executables myself.

cr1901:
Any MSDOS from 2.x on should work, though one is probably using 5.x or 6.x or FREEDOS if they're doing DOS development.

AFAIK, the Visual Studio configs (is that what you mean by "Repo Version"?) exist to make Ville's life easier, since CMake can be finicky.

CMake is the main build system. Makefiles/build systems under historical target special cases until we decide a better place for them. But since Visual Studio isn't historical, it doesn't make sense to put them under the historical directory :D.

cr1901:
Also a bit OT, but: Ville, I owe you an email at some point explaining where I've been.

I still check notifications for this repo and make sure WLA builds. I had plans earlier this year to try to create a better test suite. I'd like to also look at potential parser improvements. But contract work and this timeline in general going to hell means these plans were derailed for now.

vhelin:
No worries, first things first. :) I was also planning on working on WLA DX this weekend, but I had to spend this weekend working a little, have an important meeting on Monday morning with potential customers...

PR #354, opened 2021-01-11
Improve include search error messages

### LINKED ISSUE #351: Ommit "Could not open" message when file is found in another directory.
Currently, the `include_file()` function on `include_file.c` outputs the message below when a file is not found at first, but is found on another directory. The same happens for the `incbin_file()` function.

```
INCLUDE_FILE: Could not open "path/file.asm", trying "other-path/file.asm"... found.
```

This leads to lots of output during build if there are many includes where the file is found in other included dirs:

screenshot

Maybe it's better to ommit this info and warn the user only when a file is not found. Or it could be enable on verbose mode.

lhsazevedo:
Fixes #351 

## Issue

Currently when a included file isn't found in the first directory, the `include_file()` and `incbin_file()` functions output messagens even if the file is found. Example:

```
src/asm/main.asm:123: INCLUDE_FILE: Could not open "src/asm/src/data/foo.asm", trying "src/data/foo.asm"... found.
```

And even when the file isn't found, the error message isn't really helpful after @jwrayth contributiom for multiple include directories, because it only lists the last searched directory.

```
src/asm/main.asm:19: INCLUDE_FILE: Could not open "src/main/src/graphics/boxes.bin", trying "src/graphics/boxes.bin"... not found.
src/asm/main.asm:19: INCLUDE_FILE: Error opening file "src/main/src/graphics/boxes.bin".
src/asm/main.asm:19: ERROR: Couldn't parse ".INCBIN".
```

## Proposal

This PR updates both `include_file()` and `incbin_file()` to print the error message only after all directories have been search and the file could not be located. If the file is found in any of the included dirs, no output is printed.

In the case of a file not being found in any include folder, the user will receive a message listing all searched ones:

```
src/asm/main.asm:123: FIND_FILE: Could not open "src/graphics/boxes.bin", searched in the following directories:
src/
src/data/
Current directory.
src/asm/main.asm:123: ERROR: Couldn't parse ".INCBIN".
```

## Bonus
The file search code that was previously duplicated on both `include_file()` and `incbin_file()` was extracted to a separedted function. Some other simple refactoring was done to improve maintenance of the source code.
The result is actually less code than at the moment.

## Warning
Please review carefully my changes because I'm still learning C programming :)

### DIFF
--- a/include_file.c
+++ b/include_file.c
@@ -86,6 +86,28 @@ int try_open_file(char* directory, char* partial_name, FILE** out_result) {
   return SUCCEEDED;
 }
 
+
+void print_include_search_error() {
+
+  int index;
+
+  fprintf(stderr, "%s:%d: ", get_file_name(active_file_info_last->filename_id), active_file_info_last->line_current);
+  fprintf(stderr, "FIND_FILE: Could not open \"%s\", searched in the following directories:\n", full_name);
+
+  if (use_incdir == YES) {
+    for (index = 0; index < ext_incdirs.count; index++) {
+      fprintf(stderr, "%s\n", ext_incdirs.names[index]);
+    }
+  }
+
+  if (include_dir != NULL) {
+    fprintf(stderr, "%s\n", include_dir);
+  }
+
+  fprintf(stderr, "Current directory.\n");
+}
+
+
 int find_file(char *name, FILE** f) {
 
   int index;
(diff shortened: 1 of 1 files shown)


lhsazevedo:
Just pushed a new commit, converting the two new local functions to static.

vhelin:
Looks good, thanks!

lhsazevedo:
Awesome! Glad to contribute :rocket:

PR #360, opened 2021-01-19
Fix segmentation fault when input file cannot be found

lhsazevedo:
Recent changes introduced a segfault when the input file can't be found.

```
wla-z80 -o out.o wrongfile.asm
Segmentation fault (core dumped)
```

This happens because the `print_find_error()` function tries to print the active file name and line where the error occurred, from `active_file_info_last` struct, but it still NULL when the input file is searched:

https://github.com/vhelin/wla-dx/blob/ef508e4b302d5be62c95595b5d6b30edfdc14342/include_file.c#L90-L96

This PR fixes it by simply adding a condition to check if there is a active file before trying to print it's name and line.

### DIFF
--- a/include_file.c
+++ b/include_file.c
@@ -91,6 +91,11 @@ static void print_find_error(char* name) {
 
   int index;
 
+  if (!active_file_info_last) {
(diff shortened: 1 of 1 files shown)


vhelin:
Thanks!

PR #362, opened 2021-01-23
Rename global variable i

lhsazevedo:
Renames the global `i` to `g_source_pointer`. The `g_` prefix was added to make clearer that it is a global variable.


vhelin:
Thanks, this was a good one! And now that we have one global var with prefix g_ the next big thing would be to make all globals to start with g_ ... :) A long term wish.

lhsazevedo:
Glad to contribute :)
Just created [a new issue](https://github.com/vhelin/wla-dx/issues/365) for prefix task.

Would you still keep this message?
https://github.com/vhelin/wla-dx/blob/2e4c5026ac531ce78d493943b4fd449427e46816/pass_1.c#L646-L650

vhelin:
Ah yes, I saw it in the pull request and was going to remove it, but I was really hungry, went to cook some Tom Kha soup, and forgot all about it. :)

PR #361, opened 2021-01-19
Refactor argument flags parsing

lhsazevedo:
Refactor parse_flags function, making it easier to understand and maintain.

Changes:
- Use `flags[count][1] == 'f'` instead of `strcmp(flags[count], "-f")`
- Use a switch case instead of a if statement
- Extract complex specific flag handling logic to it own static functions (include, define, output)
- Use a while loop to iterate over all flags
- Removed some code and variables that proved to be unnecessary

## Statistics
Function average Cyclomatic Complexity of main.c dropped from 15.2 to 10.6 (-30%).
Function average lines of code count (without comments) of main.c dropped from 59.3 to 38.4 (-35%).

*Unchanged functions were ommited for brevity.*

**Before:**
```
 NLOC  Avg.NLOC  AvgCCN   Avg.token  function_cnt    file
--------------------------------------------------------------
  591      59.3    15.2      401.2              9    main.c

================================================
  NLOC    CCN    token  PARAM  length  location  
------------------------------------------------
   117     25      700      2     131  parse_flags@211-341@main.c
```

**After:**
```
 NLOC  Avg.NLOC  AvgCCN  Avg.token  function_cnt    file
--------------------------------------------------------------
  595      38.4    10.6      261.6            14    main.c

  NLOC    CCN   Token  Param  Length  Location  
------------------------------------------------
    13      3      85      3      15  parse_output_object_flag@210-224@main.c
    13      3      85      3      15  parse_output_library_flag@226-240@main.c
    22      6     201      3      26  parse_define_flag@242-267@main.c
    12      4      87      3      15  parse_include_flag@269-283@main.c
    48     16     197      3      60  parse_flag@285-344@main.c
    13      5      97      2      18  parse_flags@347-364@main.c
```

PR #363, opened 2021-01-23
Extract Mersenne Twister to it's own file

lhsazevedo:
As suggested in https://github.com/vhelin/wla-dx/pull/361#issuecomment-765751523

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,6 +135,8 @@ set(WLA_SRCS
     listfile.h
     printf.c
     printf.h
+    mersenne_twister.c
+    mersenne_twister.h
     )
 
 foreach(ARCH IN LISTS ARCHS)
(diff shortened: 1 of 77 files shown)


vhelin:
Nice, thanks!

vhelin:
I can confirm that AmigaOS3 (smake) building works!

lhsazevedo:
Great :rocket: 
One day I'll try AmigaOS3 on FS-UAE, never done this, only seen Amiga on demo compos.

vhelin:
One problem I have with Amiga is that I fail to redirect the "amiga" script's output to a file, it just keeps on going to console...

execute amiga > output.txt
execute amiga 2> output.txt
execute amiga *> output.txt

... all fail...

It'd be nice to collect the output to a file for easy browsing. Setting the WinUAE emulator speed to very low, waiting for something interesting to happen, and then breaking the script execution with CTRL-C is a really painful way to dig into what happens on the Amiga side...

PR #371, opened 2021-02-03
Use a better current directory message

### LINKED ISSUE #358: Find a better way to display current directory on failed include message
In #354 I've added string "Current directory" to the list of searched folders because I didn't and still don't know how to obtain the current working directory in a compatible way for all supported platforms, and how to display it as well. See below.

```
src/asm/main.asm:123: FIND_FILE: Could not open...
src/
src/data/
Current directory.
```

One option is to change it to the real path, but it can also be confusing as the other listed paths are all relative to the current working directory:

```
src/asm/main.asm:123: FIND_FILE: Could not open...
src/
src/data/
/home/user/myproject/
```

Another option is change it to `./`

```
src/asm/main.asm:123: FIND_FILE: Could not open...
src/
src/data/
./
```

lhsazevedo:
Fixes #358

Before:
```
src/asm/main.asm:123: FIND_FILE: Could not open...
src/
src/data/
Current directory.
```

After:
```
src/asm/main.asm:123: FIND_FILE: Could not open...
src/
src/data/
./ (current directory)
```

### DIFF
--- a/include_file.c
+++ b/include_file.c
@@ -106,7 +106,7 @@ static void print_find_error(char* name) {
     fprintf(stderr, "%s\n", include_dir);
   }
 
-  fprintf(stderr, "Current directory.\n");
+  fprintf(stderr, "./ (current directory)\n");
 }
 
 


vhelin:
Thanks :)

PR #381, opened 2021-02-15
Remove more g_ind and g_inz from pass_1.c

lhsazevedo:
Remove `g_ind` and `g_inz` from the following functions:
- `directive_bank()`
- `directive_dbm_dwm_dlm()`
- `directive_table()`
- `directive_row_data()`

Two new warnings were introduced:
```
pass_1.c:2377:20: warning: ‘slot’ may be used uninitialized in this function [-Wmaybe-uninitialized]
pass_1.c:2377:35: warning: ‘bank’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```
This happens because the `g_ind`/`g_inz` globals where used in an if/elseif condition without an else block. Now that they're locals, they may be used uninitialized if both conditions fail:
```c
int slot, bank;

if (x) {
  slot = 1;
  bank = 3;
} else if (y) {
  slot = 3;
  bank = 7;
}

/* Use bank and slot here */
```

### DIFF
--- a/pass_1.c
+++ b/pass_1.c
@@ -2391,10 +2391,8 @@ int directive_bank(void) {
 int directive_dbm_dwm_dlm(void) {
   
   struct macro_static *macro;
-  int result;
 
-  result = input_number();
-  if (result != INPUT_NUMBER_ADDRESS_LABEL) {
+  if (input_number() != INPUT_NUMBER_ADDRESS_LABEL) {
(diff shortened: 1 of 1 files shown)


vhelin:
Thanks! I'll take care of the warnings...

lhsazevedo:
Thank you :)

lhsazevedo:
I will add tests for these directives soon

Issue #319, opened 2020-06-13
Allow Makefile rules to generate when incbin fails


bazzinotti:
Hello Ville and co!

I have wla-dx to allow Makefile rules generation even when "incbin" files are not found. In my usage, this was useful when I generated the Makefile before generating other assets.

If this interests you, let's chat or ask me to submit the PR.

vhelin:
Sounds good for me! Just submit a PR and I'll approve it. :)

vhelin:
Ah, you already did submit the PR - I'll have a look at it.

PR #321, opened 2020-06-13
spc700 1-bit opcode // Makefile incbin

bazzinotti:
This PR provides insight behind #319 (makefile) and #320 (spc700 1-bit opcode). It should not be viewed as a polished solution, but to get the ball rolling if you're interested in these adjustments I made to wla-dx.

PR #383, opened 2021-02-18
Incbin fix

ToriningenGames:
Fixes issue #319, and a minor build issue in DOS/Windows environments.

This makes PR #321 half redundant, but it was Unix only, and the function signature has changed.

Tests pass, and my code now compiles whenever I add a new file.

As a side effect, builds succeed even if an `.INCLUDE` directive was used, generating potentially incomplete Makefile rules.
Depending on the structure of the Makefile, this may miss some dependencies in chained-include cases:
File A includes file B, which includes file C, but file B doesn't exist yet. Therefore, A depends on C, but the generated rule would neglect this. The rule generated for B wouldn't, though.

### DIFF
--- a/include_file.c
+++ b/include_file.c
@@ -13,10 +13,11 @@
 #include "printf.h"
 
 
-extern int g_ind, g_source_pointer, g_extra_definitions, g_parsed_int, g_use_incdir;
+extern int g_ind, g_source_pointer, g_extra_definitions, g_parsed_int, g_use_incdir, g_makefile_rules;
 extern char g_tmp[4096], g_error_message[sizeof(g_tmp) + MAX_NAME_LENGTH + 1 + 1024];
 extern struct ext_include_collection g_ext_incdirs;
 extern FILE *g_file_out_ptr;
+extern int generate_tmp_name(char **filename);
 
 struct incbin_file_data *g_incbin_file_data_first = NULL, *g_ifd_tmp;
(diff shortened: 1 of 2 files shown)


vhelin:
So when we don't find a file we create an empty file, but never delete it? It is a solution, but not a very nice one... :)

vhelin:
> Depending on the structure of the Makefile, this may miss some dependencies in chained-include cases:
> File A includes file B, which includes file C, but file B doesn't exist yet. Therefore, A depends on C, but the generated rule would neglect this. The rule generated for B wouldn't, though.

If file B doesn't exist, how could we know about file C? 

Anyway, I think I can at least fix the tmp file deletion pretty quickly, so thank you! :)

ToriningenGames:
Weird. I'm not seeing that file stick around on my machine.
I figured the file returned by that tmp file function got auto deleted later. Maybe this has something to do with that function only seemingly return one filename ever?

Perhaps use `tmpfile`/`tmpnam`?

vhelin:
I think you are correct about it getting auto deleted later as it was the filename for WLA's internal symbol stream... Well, after I made generate_tmp_name() to return an unique name every time it's called I got a loose tmp file. 

I haven't heard about tmpfile() before, but found this:

https://www.techonthenet.com/c_language/standard_library_functions/stdio_h/tmpfile.php

it's in ANSI C90, so we could use it, but I already created a unique tmp file name for makefile rules generation. :)

Issue #384, opened 2021-03-06
Access violation in Windows builds for some .include


maxim-zhao:
Recent refactoring of locals vs. globals missed some code behind `#if defined(MSDOS)` guards. PR coming...

lhsazevedo:
Fixed by #385?

PR #385, opened 2021-03-06
Wrong variable used as path index causes segfault, see #384

maxim-zhao:
Recent refactoring of locals vs. globals missed some code behind `#if defined(MSDOS)` guards.


lhsazevedo:
Oops, [my bad](https://github.com/vhelin/wla-dx/commit/e7c9290eb7fa03ec31582ee7d91891004e00edc9). Thanks!

vhelin:
Thanks! :) It was a big refactoring so at least I was prepared to see some such errors. If this was all then the refactoring was really well done!

PR #386, opened 2021-03-07
Fix determining source filename with more than 256 files

Drenn1:
The "source file id" was a byte in the object file, meaning there could only be up to 256 source files - at least, for any particular object file. I seem to have more than that, somehow, since I never really bothered to make multiple object files.

This limitation didn't really cause any problems with building, but it did cause segmentation faults when attempting to generate listfiles. This fixes that.


vhelin:
Thanks!

lhsazevedo:
Thanks for the executable bit :pray: 

Now we can also remove this from the linux pipeline:
https://github.com/vhelin/wla-dx/blob/f1a4bea1e303692abd1dcd3adf17a8fc3270705c/azure-pipelines-linux.yml#L17

vhelin:
I wasn't aware that you could set executable bits in Git... I did the same for run_tests.sh, removed two chmod+x lines from Azure pipelines (Linux) files, and let's see if the Linux tests still pass... :)

PR #393, opened 2021-03-19
Handle unused fread results

lhsazevedo:
See #16.

`fscanf` comes next!

In the future we can also extract this logic to function, maybe `read_whole_file`.

**Bonus**
`g_background_size` was demoted to a local variable, called `background_size`.

### DIFF
--- a/byte_tester/main.c
+++ b/byte_tester/main.c
@@ -128,7 +128,10 @@ int _read_binary_file(char *filename, int *did_we_read_data, FILE *f, int *file_
     return FAILED;
   }
 
-  fread(g_binary_file, 1, *file_size, fb);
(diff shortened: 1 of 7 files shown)


vhelin:
Looks good, but enhancement idea:

Also print an error message if the fread() fails. In the case of include_file.c it could be like

      snprintf(g_error_message, sizeof(g_error_message), "Could not read all %d bytes of \"%s\"!", file_size, g_full_name);
      print_error(g_error_message, ERROR_INC);
      return FAILED;

(I hope the error message works, I wrote it here without testing it)

vhelin:
> In the future we can also extract this logic to function, maybe read_whole_file.

That'd be great! :)

vhelin:
Nice, thank you!

Issue #364, opened 2021-01-24
Automate historical makefiles generation


lhsazevedo:
Here is a simple test [python script](https://gist.github.com/lhsazevedo/28c43d9e8cd534275aea90d90ad858f5#file-gen-py) for generating historical makefiles using [a template](https://gist.github.com/lhsazevedo/28c43d9e8cd534275aea90d90ad858f5#file-amiga-template). Only amiga makefiles for now, but should be easy to add msdos and smake.

It still needs some adjustments to output makefiles equivalent to the current ones, see [the diff](https://gist.github.com/lhsazevedo/28c43d9e8cd534275aea90d90ad858f5#file-diff-diff).

I'll get back to this soon.

lhsazevedo:
May help #356

cr1901:
We already require Python to generate docs. I'm fine with this change- it can be the CMake equivalent for the historical architectures. 

Maybe this script could even evolve the ability to zip up the contents for xfer to an old machine (Ville, what do you do to xfer the source tree to Amiga, seeing that I doubt git runs there)?

vhelin:
This sounds like a great idea!

@cr1901 - I sold my Amiga as I wasn't using it much, using WinUAE now.

vhelin:
Also, having such script in GitHub would be super nice, but we still need to keep on having the (generated) historical makefiles there as well as for example people who use Amiga/MSDOS might not have python executables etc...

lhsazevedo:
We can add a step to the CI pipeline that, for every commit, run the script and commit the new makefiles if there are changes. Something like:
```sh
python script.py

if git diff --no-patch --exit-code; then
  git commit -am "Update historical makefiles"
  git push
fi
```

## Finishing the script

Now I just need to figure out the patterns to the object dependencies, if there is one.
Each object depends just on i'ts own source and headers, defines.h, and makefile:
```makefile
main.o: main.c defines.h main.h makefile
parse.o: parse.c defines.h parse.h makefile
include_file.o: include_file.c defines.h include_file.h makefile
```
But printf.o does not depends on defines.h
```makefile
printf.o: printf.c printf.h makefile
```

Opcodes objects depends only on the .c source:
```makefile
opcodes_65c02_tables.o: opcodes_65c02_tables.c
opcodes_65c02.o: opcodes_65c02.c
```

Would removing defines.h from vendor objects be a problem? If not, we can use tree groups for generation:
- WLA objects (main, parse, pass_1 etc):
[...]

Issue #356, opened 2021-01-11
Move active codebase out of root directory


lhsazevedo:
What do you think of moving the active codebase (wla-cpu, wlalink, byte-tester, etc) out of the root directory?
I think this way it's simpler to understand, navigate and maintain it. As well as keeping consistent with already separated folders (docs, examples, etc)

We can go even further and create a subdir for wla-cpu inside src/, and reorganize wla-cpu code in folders too (like decode/ and opcodes/):

```
📁 examples
📁 historical
📁 src
├ 📁 wlab
│ ├ 📄 main.c
│ └ ...
│
├ 📁 wlalink
│ ├ 📄 main.c
│ └ ...
│
└ 📁 wla-cpu
   ├ 📁 opcodes
   ├ 📁 decode
   ├ 📄 main.c
       ...
```

I've setup a branch so you can clone and test it: https://github.com/lhsazevedo/wla-dx/tree/move-to-src
It builds and passes all tests here at my Linux desktop. Don't know about Windows though.

cr1901:
It's a good idea, but I won't have time to test this on Windows or MSDOS (don't forget to update those Makefiles) for a bit.

In addition, changing the directory structure is going to make bisecting more difficult. But I suppose one could write a script for that to automate rebuilding against the old dir structure and new one.

lhsazevedo:
Sadly I can't test on other platforms, and it maybe be better to let more experienced people handle Makefiles as they are new to me to. I think this is more an idea/starting point than the real thing.

Oh, and what does bisecting means in this context?

cr1901:
By bisecting I mean [`git bisect`](https://www.metaltoad.com/blog/beginners-guide-git-bisect-process-elimination).

vhelin:
Good idea as the current folder structure is a bit messy. But like @cr1901 said, the historical makefiles/scripts need to be updated as well and work after the change. Personally I wouldn't do this unless I had nothing else to do as bug fixes & new features >>> reorganization... It'll be a pain in the behind to update those files in the historical folder... Let's keep this in mind unless someone really feels like doing it!

lhsazevedo:
Maybe we can consider this now that we have the makefile generator

PR #394, opened 2021-03-21
Use automatic dependency tracking on MS-DOS makefiles

lhsazevedo:
## Intro

If accepted, this pull request will update MS-DOS makefiles adding automatic dependency tracking.
Additionally, it removes the `-m` flag (removed from gcc) in favor or `-mtune`. It also defines `NDEBUG` so we have release builds as default.

Following https://github.com/vhelin/wla-dx/issues/364#issuecomment-774841557, I was able to *successfully* build WLA 6502, 6510, 6800, 6801, 6809, 8008, 8080, gb and z80.
As before, 65816, 65c02, 65ce02, huc6280, spc700 were not built. I suspect that the target name length plays a role here as all of them result in a .exe that does not comply with the DOS 8dot3 filename limit. Further analysis is needed.

The makefiles were re-generated with a Python script. We can do the same for Amiga make/smake ones allowing #356 to proceed.

### Pattern Rules

The first step was to use just one Pattern Rule for all object dependencies, so instead of:

```makefile
all: $(OFILES) makefile
    $(LD) $(LDFLAGS) $(OFILES) -o wla-z80.exe

main.o: main.c defines.h main.h makefile
    $(CC) $(CFLAGS) main.c

printf.o: printf.c printf.h makefile
    $(CC) $(CFLAGS) printf.c

# rules omitted for brevity ...
```
We use:

```makefile
all: $(OFILES)
    $(LD) $(LDFLAGS) $^ -o wla-z80.exe

%.o: %.c
    $(CC) $(CFLAGS) $<

# "$<" means first prerequisite name
# "$^" means all prerequisites
```
The `makefile` prerequisite isn't needed and was removed.

Note that headers were removed from the prerequisites, as gcc will generate this for us below.

### -MMD -MF

By passing `-MMD -MF` options, we ask gcc to output the dependencies of the current source in makefile format to a file.

```makefile
# stack.d

stack.o: stack.c defines.h hashmap.h shared.h parse.h pass_1.h stack.h \
 include_file.h printf.h
```

Instead of manually typing (and tracking), we can just include those *.d makefiles.

So, we add `-MMD and -MF <target>.d` to our Pattern Rule:
```makefile
# "@" gets replaced by the target name.
# ".o=.d" replaces the .o extension with .d

%.o: %.c
    $(CC) $(CFLAGS) $< -MMD -MF $(@:.o=.d)
```

The effective rule for a given object is the following:

```makefile
# This is just a demonstration of the effective rule for the
# main.o target, and is not present on the makefile itself.

stack.o: stack.c
    $(CC) $(CFLAGS) stack.c -MMD -MF stack.d  
```

Instead of including each .d makefile, i'm using a little trick to include all of them:
(See the referenced links for more info)
```makefile
-include($(patsubst %.o,%.d,$(OFILES)))
```
Note that a "-" is used before the include to suppress errors, as there are no *.d files to include on the first run.
This isn't a problem because they're useful only when rebuilding objects after changing files during development.

## Dependency graph
Before:
Before
After:
After

## References
[Makefile Tutorial - Advanced | Automatic Dependency Tracking, Variables, and Pattern Rules](https://www.youtube.com/watch?v=PiFUuQqW-v8)
[Makefile Tutorial By Example](https://makefiletutorial.com)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,6 +6,7 @@
 *.md    text
 *.sh    text eol=lf
 *.bat   text eol=crlf
+makefile.msdos.*    text eol=crlf
 
 # bat-files usually works with LF, but Windows users might have a better
 # editing time with notepad and CRLF, because nodepad is dumb with LF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ wla_[ab].tmp
 *.so
 *.a
 *.exe
+*.d
 
 # Generated files
 opcodes_*_tables.c
--- a/historical/makefiles/makefile.msdos.6502
+++ b/historical/makefiles/makefile.msdos.6502
@@ -1,65 +1,22 @@
 
 CC = gcc
-CFLAGS = -m486 -c -O3 -ansi -pedantic -Wall -DMSDOS -DMCS6502
+CFLAGS = -mtune=i486 -c -O3 -ansi -pedantic -Wall -DMSDOS -DMCS6502 -DNDEBUG=1
 LD = gcc
 LDFLAGS = -lm
 WLAFLAGS = $(CFLAGS)
 
-CFILES = main.c parse.c include_file.c pass_1.c pass_2.c pass_3.c pass_4.c stack.c listfile.c crc32.c hashmap.c printf.c mersenne_twister.c opcodes_6502_tables.c opcodes_6502.c
-HFILES = main.h parse.h include_file.h pass_1.h pass_2.h pass_3.h pass_4.h stack.h listfile.h crc32.h hashmap.h printf.h mersenne_twister.h defines.h
-OFILES = main.o parse.o include_file.o pass_1.o pass_2.o pass_3.o pass_4.o stack.o listfile.o crc32.o hashmap.o printf.o mersenne_twister.o opcodes_6502_tables.o opcodes_6502.o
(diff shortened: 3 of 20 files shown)


cr1901:
IIRC, the Makefiles explicitly wrote out all dependencies b/c I couldn't rely on a particular `make` flavor being available on an MSDOS system. So I targeted the lowest-common denominator. But since these Makefiles are specific to DJGPP anyway, I guess it's not a big deal to use GNU Make features.

lhsazevedo:
Hmm, maybe we can get rid of the Pattern Rules and tweak the Python script to generate one rule for each object. The `$^`, `$<` and `$@` aren't really needed and can be replaced in the script as well. What do you think?

I guess the only important make feature is to be able to include other makefiles with `include` for the .d files. Is this supported on old versions?

lhsazevedo:
Here is a generated makefile without the features mentioned above:
https://gist.github.com/lhsazevedo/2ba850be6317362f9edd4f7b90811280

I'm looking forward to your feedback!

cr1901:
Well, even `-include` is a GNUism :P. I think your original idea is fine for building w/ DJGPP, just I don't think that the Python generator should rely on any particular `make` flavor.

(Do you intend to commit the Python script itself?)

lhsazevedo:
You are right, and there are not so many source files. I think this is a lot of unnecessary effort.
It might be better to just keep the dependencies in the script itself. See this replacement PR #399.

> Do you intend to commit the Python script itself?

Sure!

Issue #368, opened 2021-01-30
Packed ASCII Support


StephanSchmidt:
Make .ASC support packed ASCII where characters are encoded into 6bits. 

This saves 25% of storage - with the downside of only uppercase characters.

Before printing a string one would need to decode this to 8bits.

vhelin:
Taking  lowercase letters away from ASCII -> doesn't really fit into 6 bits. Something else also has to go... Any links to specifications of this 6-bit version of ASCII? Is there a 6-bit ASCII standard in old games?

I found this: https://en.wikipedia.org/wiki/Six-bit_character_code - SixBit ASCII?

StephanSchmidt:
You are right, more has to go. And there are different encodings, e.g.

https://programmingpraxis.com/2014/05/06/packed-ascii/

which might be DEC SIXBIT from your link.

Kroc:
This is too specific for the assembler; compression should be done via a python script or the like and .INCluded. (There are literally infinite number of encoding schemes people might want). There's also the file-parsing macro that runs a macro for each byte in a file. 

What would be useful, in a generic sense, that you could use to achieve packed ASCII, is the ability to output less than 8-bits at a time (e.g. `.BITS ...`?)

StephanSchmidt:
It's much more difficult to include a packed string file than packed ASCII in the source file with address labels for each string.

But yes if this is not included I'd used some python script to generate the packed ASCII and generate an include file with the labels.

vhelin:
WLA already contains some directives that make life easier so adding one more should be ok, as long as it's really useful. :)

.BITSMAP 4 "0123456789" ; takes the number of bits the mapping uses and a mappings array?
.BITS "668275" ; maps the string using the .BITSMAP into 4 bits per character?

Quite doable, if this is ok and usable?

StephanSchmidt:
Something that way would be nice, indeed.

I currently read about how Infocom did text storage for their text adventures.

They use something they have called ZSCCI which encodes lower case characters with 5 bits, and upper case characters with more bits, and punctuation with even more.

This can lead to up to a 50% reduction in memory usage for texts (but is more complicated to encode/decode), though the open source Z-machine virtual machines should have code.

https://www.inform-fiction.org/zmachine/standards/z1point0/sect03.html

vhelin:
[...]

PR #399, opened 2021-03-22
Add historical makefile generator

lhsazevedo:
*This PR is a replacement to #394. The generated makefiles do not rely on make extensions and flavors.*

If accepted, this pull request will add an makefile generator python script, and update the makefiles with the generated ones.
For now, only MS-DOS makefiles can be generated but we can add other templates afterwards.

The ideia here is to keep track of the dependencies only on the generator, instead of maintaining each an every historical makefile.

Notable makefile changes:
- `-m` flag (removed from gcc) in favor of `-mtune`.
- `NDEBUG` define added so we have release builds as default.
- Objects now have accurate prerequisites

Additionally, `printf.c` was updated to allow MS-DOS builds.

I was able to *successfully* build WLA 6502, 6510, 6800, 6801, 6809, 8008, 8080, gb and z80.
As before, 65816, 65c02, 65ce02, huc6280, spc700 were not built. I suspect that the target name length plays a role here as all of them result in a .exe that does not comply with the DOS 8dot3 filename limit. Further analysis is needed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,6 +6,7 @@
 *.md    text
 *.sh    text eol=lf
 *.bat   text eol=crlf
+makefile.msdos.*    text eol=crlf
 
 # bat-files usually works with LF, but Windows users might have a better
 # editing time with notepad and CRLF, because nodepad is dumb with LF
--- a/historical/makefile_generator/gen.py
+++ b/historical/makefile_generator/gen.py
@@ -25,6 +25,13 @@ cpus = [
     {'name': 'z80', 'define': 'Z80'}
 ]
 
+warning = [
+    '###################################################',
+    '# WARNING: This file was automatically generated. #',
+    '###################################################',
+    '',
+]
+
 # First we instantiate a new Template for each platform.
 # For now, only MS-DOS is being generated.
 msdos_stub = open('makefile.msdos.template')
@@ -182,5 +189,5 @@ for cpu in cpus:
         os.mkdir(output_folder)
 
     f = open(output_folder + '/makefile.msdos.' + cpu['name'], "w")
-    f.write(msdos_out)
+    f.write('\r\n'.join(warning) + msdos_out)
     f.close()
--- a/historical/makefile_generator/gen.py
+++ b/historical/makefile_generator/gen.py
@@ -27,7 +27,7 @@ cpus = [
 
 # First we instantiate a new Template for each platform.
 # For now, only MS-DOS is being generated.
-msdos_stub = open('msdos.template')
+msdos_stub = open('makefile.msdos.template')
 msdos_stub_contents = msdos_stub.read()
 msdos_stub.close()
 msdos_template = MyTemplate(msdos_stub_contents)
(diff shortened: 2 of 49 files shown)


cr1901:
I'm happy w/ this/will accept :). It's a good start for getting rid of the `historical` directory eventually :D!

vhelin:
I love the idea! I'll look into this later this week with better time, currently working on .BITS (define 1-32 bits) support, to fix GitHub issue 368: https://github.com/vhelin/wla-dx/issues/368

lhsazevedo:
Thanks!

I took a closer look on the failed builds they built just fine, it's the move command on msdos.bat that was failing because of the long file name. Everything OK on the WLA side.

The tests that I did were on DOSBox-X and QEMU running FreeDOS and MS-DOS 7.1. I guess it's better to let someone with access to a real DOS setup fix this later.

lhsazevedo:
Added the Amiga gcc template as well :)

I did not add the smake template because their main target defines all files (source, objects and headers) as prerequisistes, instead of just objects as the other platforms do.
https://github.com/vhelin/wla-dx/blob/e86d9451b4c3528286bfa311ab967b2307d3bc8f/historical/makefiles/smake.z80#L14-L21

vhelin:
> I took a closer look on the failed builds they built just fine, it's the move command on msdos.bat that was failing because of the long file name. Everything OK on the WLA side.
> 
> The tests that I did were on DOSBox-X and QEMU running FreeDOS and MS-DOS 7.1. I guess it's better to let someone with access to a real DOS setup fix this later.

You mean the move command that moves the .exe to binaries\ ? Could the executables be named differently (with less letters) on MSDOS? It doesn't sound like a big change in makefiles... ?

move wla-6510.exe binaries\

? Instead of "wla-*" could the executables be like "w-*" ?

vhelin:
> Added the Amiga gcc template as well :)
> 
> I did not add the smake template because their main target defines all files (source, objects and headers) as prerequisistes, instead of just objects as the other platforms do.

The Amiga GCC Makefiles are for new "Amigas", and I do want to keep supporting the Classic Motorola 680x0 Amigas and for that we use smake-files. Would it be possible to add new code to the Python script that would create then those smake-files?

PS. I'm happy to accept this PR, just say when you think everything is ready.

lhsazevedo:
> Would it be possible to add new code to the Python script that would create then those smake-files?

Yeah, sure. I'll try to setup an AmigaOS environment on FS-UAE to test the generated makefiles.

> PS. I'm happy to accept this PR, just say when you think everything is ready.

Great, will do!

vhelin:
If you cannot get FS-UAE etc. to work, just try to add as working smake-file support there as possible and I'll fix it to work if there's something to fix. :)

lhsazevedo:
Oh great, thanks, I won't be able to test it this weekend...

I just added smake support, it's ready to merge, if approved.

vhelin:
Thanks!

vhelin:
The template was missing one special object file: LIB:c.o but after I added that to the template and regenerated the smake-files they worked in WinUAE!

lhsazevedo:
Awesome :rocket:

PR #402, opened 2021-04-13
Remove most of g_ind and g_inz usage from pass_1.c

lhsazevedo:
Turn `g_ind` and `g_inz` references in pass_1.c into local variables where possible.
I also moved a few variables to inner scopes when I had the chance, but this is not the focus of this PR.

The globals were kept on the following functions as they may rely on global state.
- `parse_push_pull_registers`
- `parse_exg_tfr_registers`
- `parse_tiny_int`
- `evaluate_token`
- `parse_dstruct_entry`
- `directive_dstruct`

Although being simple, the changes are quite extensive, and I think its better if we review each commit individually.

## Stats
References to `g_ind` and `g_inz` in pass_1.c:
**Before**: 470 references.
**After**: 44 references.

### DIFF
--- a/pass_1.c
+++ b/pass_1.c
@@ -8614,7 +8614,7 @@ int directive_snesemuvector(void) {
 
 int directive_print(void) {
 
-  int get_value, value_type;
+  int get_value, value_type, number_result;
 
   while (1) {
     get_value = NO;
@@ -8635,9 +8635,9 @@ int directive_print(void) {
       get_value = YES;
     }
 
-    g_inz = input_number();
+    number_result = input_number();
 
-    if (g_inz == INPUT_NUMBER_STRING || g_inz == INPUT_NUMBER_ADDRESS_LABEL) {
+    if (number_result == INPUT_NUMBER_STRING || number_result == INPUT_NUMBER_ADDRESS_LABEL) {
       char t[256];
     
       if (get_value == YES) {
@@ -8652,7 +8652,7 @@ int directive_print(void) {
         fflush(stdout);
       }
     }
-    else if (g_inz == SUCCEEDED) {
+    else if (number_result == SUCCEEDED) {
       if (g_quiet == NO) {
         if (value_type == 0)
           printf("%x", g_parsed_int);
@@ -8661,7 +8661,7 @@ int directive_print(void) {
         fflush(stdout);
       }
     }
-    else if (g_inz == INPUT_NUMBER_EOL) {
+    else if (number_result == INPUT_NUMBER_EOL) {
       next_line();
       break;
     }
--- a/pass_1.c
+++ b/pass_1.c
@@ -5579,14 +5579,14 @@ int directive_rombanks(void) {
 
 int directive_rombankmap(void) {
   
-  int b = 0, a = 0, bt = 0, bt_defined = 0, x, bs = 0, bs_defined = 0, o, q;
+  int b = 0, a = 0, bt = 0, bt_defined = 0, x, bs = 0, bs_defined = 0, o, q, token_result;
 
   no_library_files(".ROMBANKMAP");
 
   /* ROMBANKMAP has been defined previously */
   if (g_rombankmap_defined != 0 || g_rombanks_defined != 0) {
     o = 0;
-    while ((g_ind = get_next_token()) == SUCCEEDED) {
+    while ((token_result = get_next_token()) == SUCCEEDED) {
       /* .IF directive? */
       if (g_tmp[0] == '.') {
         q = parse_if_directive();
@@ -5678,7 +5678,7 @@ int directive_rombankmap(void) {
         }
       }
       else {
-        g_ind = FAILED;
+        token_result = FAILED;
         break;
       }
     }
@@ -5686,7 +5686,7 @@ int directive_rombankmap(void) {
   /* no ROMBANKMAP has been defined */
   else {
     o = 0;
-    while ((g_ind = get_next_token()) == SUCCEEDED) {
+    while ((token_result = get_next_token()) == SUCCEEDED) {
       /* .IF directive? */
       if (g_tmp[0] == '.') {
         q = parse_if_directive();
@@ -5765,13 +5765,13 @@ int directive_rombankmap(void) {
         }
       }
       else {
-        g_ind = FAILED;
+        token_result = FAILED;
         break;
       }
     }
   }
 
-  if (g_ind != SUCCEEDED) {
+  if (token_result != SUCCEEDED) {
     print_error("Error in .ROMBANKMAP data structure.\n", ERROR_DIR);
     return FAILED;
   }
--- a/pass_1.c
+++ b/pass_1.c
@@ -5463,7 +5463,7 @@ int directive_name_gb(void) {
 
 int directive_rombanks(void) {
 
-  int q;
+  int i, q, bank_address;
 
   no_library_files(".ROMBANKS");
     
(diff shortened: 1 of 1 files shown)


vhelin:
A question about how you did this (good work, BTW!): Did you remove global g_ind, then add local g_ind to the functions, and then renamed g_ind inside functions?

vhelin:
@cr1901 Could you help us out a little with reviewing some of these commits? There's a lot of them...

lhsazevedo:
> Did you remove global g_ind, then add local g_ind to the functions, and then renamed g_ind inside functions?

That seems clever, but I did it the dumb way ha. I just gone through each function searching for `/g_in[dz]/`, then replaced with one or more locals with a meaningful name. Runned tests before each commit.

> good work, BTW!

Thanks! Glad to contribute. :rocket:

vhelin:
Checked so far (the first 4 in this PR):

* Remove g_inz from macro_start_dxm
* Remove g_inz from directive_db_byt_byte
* Remove g_inz from parse_enum_token
* Move enum_add_struct_fields() vars to inner scopes 

... and I'm out of energy. :) Too much bug hunting today at work. Will continue later...

cr1901:
@vhelin I skimmed the rest. These commits were easy to skim because they are mainly text subsitutions and moving variables to smaller scopes. The changes look fine to me, with my comments below:

* `directive_dsl` appears to still require global state (`g_parsed_int`)?

* In commit 0aa4a424e, is `g_ind` not used _at all_ anymore? I know it got written to a lot over many functions, but it feels weird to me that removing it from the for loop doesn't break anything.

* Commits 8c8928ff, (end of) 06a40469, and 0472fe49 looks fine to me, but I wonder if there was a reason to assign to a variable and then check the result? Is it possible that `g_inz` was consulted after `directive_section()` returned?

lhsazevedo:
Thank you!

> directive_dsl appears to still require global state (g_parsed_int)?

Yeah, pretty much all of them require global state. When calling `input_number()`, the parsed value is written to `g_parsed_int`. The returned value is just a status code. Almost all of them call `input_number()`.

But the main thing here is reducing the usage of `g_ind` and `g_inz`, two temporary globals that are being reused instead of locals. Will try to reduce global state when I get better with C! :P

> In commit 0aa4a42, is g_ind not used at all anymore? I know it got written to a lot over many functions, but it feels weird to me that removing it from the for loop doesn't break anything

I looked if it was being used inside the function itself and it was not. Now I just checked the places where `parse_dstruct_entry()` is called and yes, it appears that no data is returned via `g_ind`.

It was implemented in https://github.com/vhelin/wla-dx/commit/e996850235953fe8aeab668d598f8fafa934da1d, maybe the author @Drenn1 could confirm this.

> Commits 8c8928f, (end of) 06a4046, and 0472fe4 looks fine to me, but I wonder if there was a reason to assign to a variable and then check the result? Is it possible that g_inz was consulted after directive_section() returned?

Hmm, let me check...
Ok, these function are called in only one or two places and `g_inz` is not consulted after.
I guess it's was just copy pasted and kept that way.

vhelin:
I'm afraid that looking just the diffs isn't enough as the way this refactoring was made means that there might still be g_ind/g_inz references inside those functions marked as taken-care-of. Unless lhsazevedo can convince us that there aren't. :) Which would be great. For me to be confident that there aren't any not-taken-care-of g_ind/g_inz references I need to look at the code and search for all the places where g_ind/g_inz appear inside those functions, and compare the code with the diff to see that the diff doesn't miss anything...

Or am I just being a nitpicker? :)

cr1901:
>Yeah, pretty much all of them require global state.

What I was getting at is: perhaps you should add `directive_dsl` to your list in your initial message in the PR :P.

@vhelin To me, the changes look mostly like text substitutions. It's _possible_ that  If the test suite still passes, I'd rather commit this change and see if we get complaints about broken code. If we do a merge commit, we can still bisect later to figure out what's wrong.

Two things I'd like to do:

* `squash` this PR down to one commit in case it becomes clear we need to revert.
* Create a separate branch where all these individual commits still exist, so if breakage occurs, we can bisect to find the broken commit.

Open to other ideas of course :).

vhelin:
@cr1901 What I'm worried about are g_inz's that were accidentally left there in the functions without changing them to new variables (are there any?). As we still have global g_inz the code will compile...

cr1901:
_I_ didn't see any from skimming all the functions. Like I said, the diff was mostly text substitution. I don't recall seeing `g_inz` being left untouched. But if it was, I am not in a position to debug whether it was justified leaving it untouched (nor could I tell you which place where touching `g_inz` might not be a good idea).

lhsazevedo:
I'll address your points later tonight

vhelin:
Checked the following commits (5-36):

* Remove g_ind from directive_db_byt_byte

... and

* Remove g_ind/z from directive_hex
* Remove g_ind from directive_asctable_asciitable
* Remove g_ind from directive_asc
* Remove g_ind/z from directive_dw_word_addr
* Remove g_ind/z from directive_dl_long_faraddr
* Remove g_inz from directive_dsl
* Remove g_ind/z from directive_dd
* Remove g_inz from directive_dsd
* Remove g_ind from directive_name_w65816
* Remove g_ind from parse_dstruct_entry
* Remove g_inz from directive_dsb_ds
* Remove g_inz from directive_dsw
* Remove g_ind/z from directive_incbin
* Remove g_inz from directive_section
* Remove g_ind from directive_block
* Remove g_ind from directive_name_gb
* Remove g_ind/z from directive_rombanks
* Remove g_ind from directive_rombankmap
* Remove g_ind from directive_memorymap
* Remove g_inz from parse_if_directive
* Remove g_ind from parse_directive
* Remove g_ind from directive_printt
* Remove g_inz from directive_print
* Remove g_ind/z from directive_snesemuvector

... and

* Remove g_ind/z from directive_snesnativevector
* Remove g_ind/z from directive_snesheader
* Remove g_ind from directive_rept_repeat
* Remove g_ind from directive_macro
* Remove g_ind from directive_smsheader
* Remove g_ind from directive_undef_undefine
* Remove g_ind/z from directive_gbheader

All looked good. Took me 1h 30min with two small breaks... :) Squash + separate branch sounds good, but I'm ok with merging this PR.

lhsazevedo:
> @vhelin
> Took me 1h 30min with two small breaks...

Oh wow. On the future, what can I do to make the review process easier?

> @cr1901
> What I was getting at is: perhaps you should add directive_dsl to your list in your initial message in the PR :P

Hmm, I read it again and it was misleading indeed. What I meant is that I didn't remove g_ind/z from those functions because they rely on them, as you can see below. I just edited to make it clearer, thx.

> @vhelin
> Unless lhsazevedo can convince us that there aren't. :)

During the refactor, I ensured that noting was left behind but I double checked just in case:
On pass_1.c, there are now 44 references to `g_ind`/`g_inz` remaining on pass_1.c, of which 6 are to `g_ind` and the other 38 to `g_inz`:

**g_ind**:
\#    | lines       | function
----- | ----------- | ---
1     | 53          | (declaration)
2 - 6 | 1166 - 1316 | evaluate_token

**g_inz**:
\#      | line        | function
------- | ----------- | ---
1       | 53          | (declaration)
2 - 4   | 887 - 895   | parse_push_pull_registers
5 - 8   | 1015 - 1034 | parse_exg_tfr_registers
9 - 10  | 1068 - 1070 | parse_tiny_int
11 - 22 | 1175 - 1212 | evaluate_token
23 - 36 | 3738 - 3805 | parse_dstruct_entry
37 - 38 | 4059 - 4064 | directive_dstruct

It is worth mentioning that, now, the only place in the whole project that `g_ind` is being used is at `evaluate_token` :leaves:.

I am not able to create a new branch here, and I don't know how to squash keeping the original commits on another branch. Let me know if there is anything I can do now to help the merge process!

cr1901:
Then let's not bother squashing. If something breaks we need to be able to bisect it. And 36 commits isn't a huge amount in the grand scheme of things. Better to keep each commit small since this was an invasive change.

vhelin:
> Oh wow. On the future, what can I do to make the review process easier?

In this particular case the way you did variable renaming meant that there might be cases where some g_inz references were left unhandled by accident. If you had removed global g_inz, then added local g_inz vars to functions, and then used some IDE's renaming tool to rename the local g_inz vars/references to something better, that would have been good enough for me and I wouldn't had to recheck the functions...

Now that you say that

> During the refactor, I ensured that noting was left behind

well, perhaps this information would have been enough for me. :) But what's done is done and taken care of and I'm happy that these renamings are done and checked. Sometimes things need a bit extra effort.

vhelin:
It'd be great if something could be done to the CPU instruction parsing code that are now in files like e.g., decode_z80.c. Perhaps move the parsing loop frames from pass_1.c to those files, make it so that those files can be compiled to object files instead of including them to pass_1.c... That'd be a great start! After that renaming g_inz/g_ind in them might be easier...

lhsazevedo:
Awesome! Thank you guys for taking the time to review this. :rocket:

PR #406, opened 2021-04-24
Update Byte Tester to read tests from comments on source file

lhsazevedo:
_See #405 for the initial discussion of this feature._

If accepted, this PR will introduce a new `-s` flag in the Byte Tester, that allows reading test commands from comments in the assembly file starting with `@BT`. It also updates the Z80 mnemonics test to use this new mode.

In the future, we can improve this feature by allowing multiple source files to be passed.


byte_tester/main.c
```diff
@@ -140,15 +140,89 @@ int _read_binary_file(char *filename, int *did_we_read_data, FILE *f, int *file_
...
 
   char tmp[256], test_id[256], tag_id[256];
+  char *input_name = NULL, *input = NULL, *current = NULL;
   unsigned char bytes[256];
-  int file_size, end, byte_count, i, j, length, tag_start, tag_end, wrong_bytes, failures, use_address = NO, address = 0, did_we_read_data = NO, got_it = NO;
+  int input_size, file_size, end, byte_count, i, j, length, tag_start, tag_end, wrong_bytes, failures, should_extract_comments = NO, use_address = NO, address = 0, did_we_read_data = NO, got_it = NO;
   FILE *f;
   
-  if (argc != 2 || argv == NULL) {
+  if (argc < 2 || argv == NULL) {
```
vhelin, line 225: Why we allow e.g., 10 arguments while only "-s" and filename are used?

lhsazevedo: Hmm, you are right.

We can allow only two or three arguments:
```c
if ((argc != 2 && argc != 3) || argv == NULL) {
```

Or a range where  `1 < argc < 4`, that doesn't makes much sense right now but could be useful in the future when more args gets added:
```c
if (argc < 2 || argc > 3 || argv == NULL) {
```

Which one do you prefer?

vhelin: Both are fine, but perhaps 

if (argc < 2 || argc > 3 || argv == NULL) {

is minimally better? :)

lhsazevedo: I agree :)

Done.

vhelin:
Just realized, is it too early to delete "testsfile" here as run_tests.sh doesn't use the new byte_tester comment reading capabilities? With this PR we now have one test project less...

lhsazevedo:
I updated the test's makefile yesterday to call the Byte Tester:

https://github.com/vhelin/wla-dx/blob/1e0d88ccfa5d96cdce29161cae4e43600afd571c/tests/z80/mnemonics_test/makefile#L11-L17

The reason why I didn't updated run_tests.sh for this is: while testsfiles always have the same name, the source file name can change, or — in the future — the test could have multiple source files with instructions.

We can keep the testsfile as well just in case, but the test will run twice. Not a big problem though.

vhelin:
Ah, missed that makefile mod. Looks good then!

lhsazevedo:
Great, Thanks for taking the time to review carefully!

🚀

PR #407, opened 2021-04-24
Fixes for building under MSVC

maxim-zhao:
- printf.c: change for MSDOS needs to exclude WIN32
- stack.c: explicit cast double to int
- byte_tester/main.c: explicit cast ptr_diff_t to int


maxim-zhao:
I guess these aren't a problem for your continuous build. Here I'm building x64 with cmake/Ninja/VS2019.

vhelin:
The more platforms/configurations we can support the merrier! Thank you! :)

PR #411, opened 2021-05-05
MSVC build fixes (again)

maxim-zhao:
Ninja build seems to default to a higher warnings-as-errors level. This explicit cast makes it happy.


vhelin:
Thanks!

Issue #417, opened 2021-05-24
Address-to-line mappings should give a memory address instead of a ROM offset


SolraBizna:
Okay, I know I said it looked great, but here's ooooooooooone more of these... sorry!

The second field of an address-to-line mapping currently gives an offset from the beginning of the ROM. This is quite awkward for me to process, since in my target system the relationship between intended memory address and ROM offset can be... complicated. I may just not be thinking clearly about it, but I think in one of my actual cases it's even ambiguous\*. In addition, WLA appears to truncate the offset to 16 bits, so even if I do manage to parse the address correctly, ROMs larger than 64K will have problems. Labels, on the other hand, are currently given as memory addresses, which carry no ambiguity and make it easy for my debugger. It would be easier for me if both were given as memory addresses, not to mention appealing to my sense of consistency.

Would this be complicated to set up? For that matter, are there existing applications that this would break? How long has the address-to-line mapping feature even been here? (I only found it this weekend, minutes after wishing it existed, so I've been choosing to assume this feature sprang into being as a direct result of my wish, rather than the boring explanation that it's been there for a while and I'm just not up to date on new features.)

\*Specifically, some cartridges use something called "wide range hack" where the slot number affects which ROM chip is selected, and that chip is hooked up to the A pins in a different way. It's a long story... but if each `.sym` file consistently gives the memory addresses instead of ROM offsets, the debugger doesn't need to know about the "wide range hack" or any other screwy wiring of the cartridge in order to get it right.

vhelin:
The first field is bank (+ base)?

I'm not sure if symbol file writes has the address information when writing addr-to-line mapping, but I'll check it out today evening. 

I don't remember when this feature was added as I think someone else than me added it. :) But we can change the WLA symbol file format as this is now going to be v10.0 release, a new major number, and thus we can change some things a little.

SolraBizna:
I was calling each number a separate field, I'll call them numbers instead to avoid confusion. My testing indicates that the second number actually is the offset from the start of the ROM, not from the start of the bank (which might have been originally intended?).

It looks like @CypherSignal added this feature originally. If they added it, they probably needed it, and if they need it, they might be annoyed if it changes now. (Thoughts, @CypherSignal?) I'd guess it was probably mainly intended to target the SNES, whose ROM address mappings are pretty straightforward...

> But we can change the WLA symbol file format as this is now going to be v10.0 release, a new major number, and thus we can change some things a little.

[...]

PR #418, opened 2021-05-25
Update symbols.rst to include version history

CypherSignal:
Change for discussion at https://github.com/vhelin/wla-dx/issues/417#issuecomment-847331286_


vhelin:
Nice, thank you!

Issue #422, opened 2021-07-04
Convert tests to the new in-source tests instructions feature


lhsazevedo:
Keeping the test instructions inside the source file greatly reduce the amount of context switching when authoring and reviewing tests.

Recently, the Byte Tester gained the ability to read the tests instructions from the sources itself using the `-s` flag. The [z80 mnemonics_test](https://github.com/vhelin/wla-dx/blob/master/tests/z80/mnemonics_test/main.s) already uses this feature. 

If you are OK with this, I can convert the remaining tests to this new mode.
Note that tests that checks the output of multiple source files will be converted after #421 is implemented.

vhelin:
I'm ok with this! :)

vhelin:
My commit here https://github.com/vhelin/wla-dx/commit/08a494f1aaea05d37aab04b3897092b5613b914d converted the last tests so we can close this issue.

lhsazevedo:
Nice! I hope this made writing tests less boring

vhelin:
It's convenient to have the test data right next to the code it tests, but my main motivation for converting the last tests was that on Amiga my test runner script cannot use testfiles, but the embedded tests work just fine. Now that I was having trouble with the Amiga build I wanted to run all the tests on Amiga that I could...

PR #423, opened 2021-07-06
Convert Z80 tests to in-source instructions

lhsazevedo:
See #422


tests/z80/linker_test_1/main.s
```diff
@@ -273,6 +274,10 @@ _loop:
 .printv 100
 .printt "\n"
 
+// BT main.o
```
vhelin, line 277: Does this mean that we are missing these two tests that check if a definition exists or not?

vhelin (approved):
Thanks!

PR #432, opened 2021-08-29
wlalink: fix build error

rofl0r:
apparently nobody compiled with DEBUG mode on for a long time...


vhelin:
Thanks!

vhelin:
> thanks for merging. maybe might make sense to toggle debug option on in one CI job ?

Now that you mention it, it's actually a really good idea. :) I'll try to do that...

PR #434, opened 2021-09-22
SRAMSIZE max value changed to 7

KhazWolf:
SRAMSIZE could perhaps be even higher but the maximum should be at least 7


vhelin:
Good find! I have no memory of why it was set to 3, perhaps I was using an old tech document when adding support for SRAMSIZE for SNES... ? Thanks!

Issue #446, opened 2021-12-15
Makefile generation issues


maxim-zhao:
I have been using makefile generation to improve my build process but I have found a few issues... here's an example input:

```
.memorymap
defaultslot 0
slot 0 $0000 $4000 ; ROM
.endme

.rombankmap
bankstotal 1
banksize $4000
banks 1
.endro

.block "foo"
.db 0
.endb

.include "foo\bar.asm"
.include "foo\t1.asm"
.include "foo/bat.asm"
.include "foo/t2.asm"

.incbin "foo\bar.bin"
.incbin "foo\t1.bin"
.incbin "foo/bat.bin"
.incbin "foo/t2.bin"
```
Resulting in:
```
INTERNAL_PASS_1: Block "foo" is 1 bytes in size.
foo.o: test.asm \
	foo\bar.asm \
	foo	1.asm \
	foo/bat.asm \
	foo/t2.asm \
	foo\bar.bin \
	foo	1.bin \
	foo\bat.bin \
	foo\t2.bin
```

1. Block sizes go to stdout
Thus the makefile is not valid.

2. "\t" in paths is converted to a tab
Presumably other control characters get parsed too. So maybe I can use forward slashes throughout as a workaround...

3. "/" is converted to "\\" for `.incbin` but not `.include`, if I run on Windows
This inconsistency gets confusing, Now I need to make sure my makefile patterns work for both types of slash.

Not in the example...

4. Some syntax using external files doesn't play properly with this mode

For example `.stringmaptable`, which I added. Example syntax:

`.stringmaptable foo bar.tbl`

If bar.tbl does not exist, the build fails. If it does, it is not emitted to the makefile list. Can you offer some guidance on how to support this?

Here's what I think happens with the five directives I can think of that load files:

| Directive | If file not present | If file is present |
|----------|-------------------|------------|
|`.fopen` | Build fails | Not in makefile|
|`.background` | Build fails | In makefile|
|`.stringmaptable` | Build fails | Not in makefile|
|`.include` | Build passes, in makefile | In makefile |
|`.incbin` | Build passes, in makefile | In makefile |

vhelin:
Too busy at the moment to work on WLA DX, will have more free time 24-31st this month, will return to this then...

maxim-zhao:
My immediate fix was to use forward slashes throughout (asm and makefile), and have a post-processing step on the makefile generation to filter and correct the slashes; Python code:
```python
def fix_makefile(path):
[...]

PR #448, opened 2021-12-16
Do not print block sizes in quiet mode.

maxim-zhao:
makefile mode already enables quiet mode. This makes block size printing be suppressed by quiet mode.
This relates to #446


vhelin:
Thanks!

PR #449, opened 2021-12-16
clang-format file

maxim-zhao:
Not a great deal of thought here but adding this allows clang-format to tidy the code, and for my use case, allows Resharper C++ to autoformat things to match WLA DX's slightly unusual style.


vhelin:
Thanks!

PR #451, opened 2021-12-16
Makefile generation fixes for .stringmaptable

maxim-zhao:
* Emit stringmaptable filenames to makefile
* Do not fail the build when the file is not found, if in makefile mode

Relates to #446


maxim-zhao:
The build failure above is erroneous (Azure warning about some infra changes), but it seems I can't trigger a rebuild.

maxim-zhao:
Fixed merge conflicts.

include_file.c
```diff
@@ -509,38 +510,37 @@ int print_file_names(void) {
 
   struct incbin_file_data *ifd;
   struct file_name_info *fni;
-
+  struct stringmaptable *smt;
+  int is_first_line = YES;
```
maxim-zhao, line 514: Note that I changed the way this method works to simplify the logic about when to add space-newline-tab.

vhelin:
Thanks!

PR #452, opened 2021-12-16
External filename processing standardisation

maxim-zhao:
This PR does a few things:
1. Adds a flag for input_number() to disable processing of embedded control characters like \t, \n
2. Changes directives that are reading a filename to disable this processing while reading the filename token (note: this means you can't have filenames with control characters in them; this is probably a good thing)
3. Adds a call to localize_path() to a few of these directives, so slashes are normalised to the host environment (note: this is unnecessary in Windows, which tolerates both ways round, but presumably other platforms are less forgiving)
4. Fixes a bug in `.include` that meant that it localised the path using the wrong string

This relates to #446.

### DIFF
--- a/parse.c
+++ b/parse.c
@@ -16,7 +16,7 @@
 int parse_string_length(char *end);
 
 int g_input_number_error_msg = YES, g_ss, g_string_size, g_input_float_mode = OFF, g_parse_floats = YES;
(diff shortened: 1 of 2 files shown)


vhelin:
Nice, thank you!

PR #461, opened 2022-01-18
byte_tester tests have the built rom as a dependency

Drenn1:
This fixes ./run_tests.sh when make is run with multiple threads (ie. when MAKEFLAGS="-j4" is an environment variable). When multiple threads are used then it would attempt to do the check before the ROM has actually compiled because it tries to both built the ROM and do the test in parallel.


vhelin:
Thanks!

lhsazevedo:
Hmm, will remember this next time, thanks.

PR #469, opened 2022-02-08
MSVC build fixes

maxim-zhao:
it does not like assigning 64-bit strlen to int without an explicit cast.


vhelin:
Oh crap I think I always forget this... :) But thanks!

PR #473, opened 2022-02-10
Change to dynamically-allocated memory for .stringmaptable strings

maxim-zhao:
As discussed at https://github.com/vhelin/wla-dx/discussions/472, here's a first step to not using MAX_NAME_LENGTH.


vhelin:
Thanks!

PR #474, opened 2022-02-10
Remove use of temporary buffers for printing error messages

maxim-zhao:
* Removing g_error_message and g_xyz
* Make `print_error` accept the formatting arguments directly
* Fixed a few places that seemed to not log correctly...

As I changed the signature to `print_error`, there's a lot of changes here. It'll also conflict with #473, I can resolve that after you merge this or that.

(Related to #472)


vhelin:
Look good, thanks!

PR #475, opened 2022-02-10
Change MAX_NAME_LENGTH logic to give the higher limit in Windows

maxim-zhao:
We need a little extra logic to enable longer strings in Windows, as for some reason it also defines MSDOS.


vhelin:
Ah, good observation! Thanks!

PR #488, opened 2022-03-02
Change definitions to use heap-allocated strings

maxim-zhao:
Part of #472

### DIFF
--- a/defines.h
+++ b/defines.h
@@ -505,8 +505,8 @@ struct optcode {
 #define DEFINITION_TYPE_ADDRESS_LABEL 3
 
 struct definition {
-  char   alias[MAX_NAME_LENGTH + 1];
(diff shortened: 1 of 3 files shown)


maxim-zhao:
Hmm, Valgrind says it is leaking, I'll have a look later...

maxim-zhao:
I think this is ready to merge now...

vhelin:
Thanks!

maxim-zhao:
Thanks for eb67746e7511d91dc9616aecb12ecad828e7473e - I had tried to do that myself but I wasn't sure about the part where it seemed to be copying substrings.

vhelin:
In the code that I modded there really should not be any substrings, I was just circulating that strlen(s->string) in a variable to avoid recalculating it later... That was perhaps just confusing and weird "speedup" and in reality pretty useless... :)

PR #492, opened 2022-03-07
Fix for build warning-as-error in NSVC (assigning size_t to int)

maxim-zhao:


vhelin:
Thanks!

And damn, another strlen() without int-cast... I really need to remember this in the future. :)

Issue #29, opened 2015-02-16
UTF-8 support


maxim-zhao:
It's sometimes restrictive to have to use ASCII, e.g. not able to use extended box drawing characters. I guess people using accented characters would find it useful too. I can imagine this would be a very painful change, though...

```
.asciitable
map '£' = 39
.enda
```

```
INPUT_NUMBER: Got 'ú' (163) when expected "'".
ERROR: Couldn't parse "map".
```

vhelin:
Oh man, making this work would require changes all over the code. :) Linker
less, assembler a lot more...

cr1901:
I don't think that can be done without making the code non-ANSI C, (which admittedly is my goal to keep the code ANSI-fied, not Ville's). Personally, I think the gain is too little for the work that is required.

If I need non-ASCII characters, I'd just make a define that describe the character:
```
.DEFINE pound_char 0x39
```

If you're meant to print the character on the target system, you'll need to create your own set of glyphs in a paint program anyway :D.

maxim-zhao:
Yes, I was just hoping for stuff I can embed in strings. Here's my use case:

https://github.com/maxim-zhao/smsgraphicboard/blob/8ca191f2c1a03abce88ada328c554f10a5b88ab8/Dasm/Sega%20Graphic%20Board%20v2_0%20%5BProto%5D.sms.asm#L2310

It's helpful for layout to have a single character inside the `.asc` map to a glyph.

Imagine using WLA to work with accented text, or Japanese - you really want to say
```
.asciitable
map 'ぁ' to 'ん' = 0
.enda
.asc "ひらがな"
```
Doing all that with definitions would be nightmarish - you'd probably go to .tbl files and third-party tools to preprocess the text.

cr1901:
Your changes would require WLA to be Unicode-aware wrt input source files. I'm not about to add a dependency on libiconv in the main branch, but maybe there can be a fork that is Unicode-aware.

In any case, this is something I have to think about/get back to you.

cr1901:
Been thinking about this some more; one of the bad things about only knowing/requiring English on a daily basis is that I'm really not qualified to implement i18n. 

Perhaps what should be done is to have a define for UNICODE support and ASCII support. Or perhaps a WLA mode for Unicode variable-length lexer/parser, and ASCII lexer/parser (and again, just make sure that Windows users save as UTF-8).

[...]

PR #494, opened 2022-03-07
UTF-8 support for labels without : and definition names

maxim-zhao:
Cast opcode chars to unsigned to enable correct handling of labels starting with extended characters with no colon

Relates to #29

### DIFF
--- a/parse.c
+++ b/parse.c
@@ -1363,7 +1363,7 @@ int get_next_plain_string(void) {
       g_ss++;
       g_source_pointer++;
     }
-    else if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '.' || c == '{' || c == '}' || c == '\\' || c == '@' || c == ':') {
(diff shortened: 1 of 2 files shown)


parse.c
```diff
@@ -1363,7 +1363,7 @@ int get_next_plain_string(void) {
       g_ss++;
       g_source_pointer++;
     }
-    else if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '.' || c == '{' || c == '}' || c == '\\' || c == '@' || c == ':') {
+    else if ((unsigned char)c > 127 || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '.' || c == '{' || c == '}' || c == '\\' || c == '@' || c == ':') {
```
vhelin, line 1366: Please add a comment there telling people what's going on. And perhaps make c by default to be unsigned char? :) Otherwise loving how simple this fix is!

vhelin: On second thought I think I can polish this myself... :)

PR #504, opened 2022-05-14
Fixes for MSVC compiler

maxim-zhao:
Yet more strlen casts to int...

### DIFF
--- a/pass_3.c
+++ b/pass_3.c
@@ -1132,7 +1132,7 @@ int is_label_anonymous(char *label) {
   if (c == '-' || c == '+')
     return YES;
 
-  length = strlen(label);
+  length = (int)strlen(label);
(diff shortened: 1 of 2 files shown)


vhelin:
Sorry, and thanks!

maxim-zhao:
Could you add an msvc build to your continuous builds? I'm just doing this on my fork (in an MSVC AppVeyor environment):
```
cmake .
cmake --build . --config Release -j
```
I've no experience with the Azure tools for this...

vhelin:
I managed to create a new Azure Pipeline for Windows+MSVC, but couldn't run the test Bash scripts there. I think we need a .bat version of run_tests.sh

vhelin:
Does one build the Debug version like this?

```
cmake .
cmake --build . --config Debug -j
```

PR #539, opened 2022-10-24
Fixes for makefile generation:

maxim-zhao:
* We only really support GNU Make so we have to use forward slashes for directory separators
* Anything else written to stdout will break the file. So enable quiet mode in makefile mode.

### DIFF
--- a/include.c
+++ b/include.c
@@ -504,6 +504,18 @@ char *get_file_name(int id) {
   return get_file_name_error;
 }
 
+void print_file_name(FILE *f, char *prefix, char *file_name) {
+  /* Converts filename to forward slashes for make compatibility */
+  fprintf(f, "%s", prefix);
+  for (char c = *file_name++; c != 0; c = *file_name++) {
+    if (c == '\\') {
+      fputc('/', f);
+    }
+    else {
+      fputc(c, f);
+    }
+  }
+}
 
 int print_file_names(void) {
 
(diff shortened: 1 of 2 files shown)


vhelin:
Thanks, looks good! I'll clean up the style a bit and update CHANGELOG.

Also, what is this? :)

-D WLA_DX_BUG_WORKAROUND

Is there a bug I could fix?

maxim-zhao:
It relates to some definition maths which fails with this:
```
INPUT_ERROR: .IF needs immediate data.
```
...because the value is dependent on .include statements which have silently failed. So it's not so much a bug workaround as a circular dependency - I want to generate the makefile to trigger the creation of the `.include`d file, but to do so I have to add logic to allow it to complete successfully.

vhelin:
Ooh, ok, sounds complex. But if you can think of a way WLA could handle this situation, if it can be common, perhaps via a flag on the command line or by other means, feel free to open an issue here. If that feature could be useful to you, it could be useful to others and thus worth the implementation...

maxim-zhao:
It might be solvable in my specific case if we had a built-in max() function, but only if that calculation could be deferred to link time. There will always be unsolvable situations though.

vhelin:
Ok so a built-in, delayable function max(a, b). If/when that's implemented min(a, b) comes almost free with the same effort. Sounds like a feature request! I'll add an issue about it!

PR #553, opened 2023-01-31
decode.c no longer uses g_ind and g_inz

jeffythedragonslayer:
first pass at cleaning this up - I just did a rename in Notepad++ so the global vars are no longer used.  Is there a way to run all the tests in one go?

### DIFF
--- a/decode.c
+++ b/decode.c
@@ -41,8 +41,7 @@ extern int g_use_wdc_standard;
 extern int g_input_number_expects_dot;
 #endif
 
-/* NOTE: get rid of these two! */
-extern int g_inz, g_ind;
+static int s_inz, s_ind;
 
 static struct instruction *g_instruction_tmp;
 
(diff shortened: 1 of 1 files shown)


vhelin:
Well, this pull request has run Azure Pipelines CI which in turn run run_tests.bsh which you can find in the root folder (that runs all the tests in one go). Seems everything is ok, but I'll take a look today evening just in case...

vhelin:
This pull request doesn't technically change anything big enough, we still use vars defined outside the functions, this pull request makes them only local to decode.c, not local to functions. It's a start, though...

vhelin:
Then again it's better than nothing

vhelin:
There are actually no unit tests in WLA DX, only system level tests that run_tests.bsh/sh run. But on theory level I cannot see why this pull request would have broken anything. As the tests pass I'm happy. :)

PR #554, opened 2023-02-03
exclude auto-generated makefile from git

jeffythedragonslayer:
This makefile shouldn't be under version control because it is auto-generated.


vhelin:
Thanks!

Issue #23, opened 2014-11-30
Refactoring


cr1901:
This is a long term goal. I think we can all agree- WLA needs a refactoring, to make the code easier to parse (As well as for me to finish the "adding a new architecture" guide :P).

This is a long term, not-at-all trivial goal. But I think it should be possible in baby steps. After learning about what happened to Netscape, I am convinced that a full rewrite is never a good idea, so changes would need to be made to preserve the old, working code while the new code is developed/tested, and eventually supplants the older code. 

I think the first step would be to global-rename a number of global vars, and then trying to remove some of the global state. I think some vars, like the index into the current file (i), should stay global for now. There are likely other vars which make sense that other parts of the assembler should be able to modify them.

cr1901:
I haven't really been work on WLA. since September, I don't really have much time, and WLA still really needs to be refactored. I've found it difficult to flush out parser bugs; to my knowledge, WLA's macro language is not context-free.

I suspect parsing rules being applied incorrectly is the biggest source of bugs. Did you ever write a grammar for WLA?

vhelin:
On 19 Apr 2016 10:18, "William D. Jones" <notifications@github.com> wrote:
>
> I haven't really been work on WLA. since September, I don't really have
much time, and WLA still really needs to be refactored. I've found it
difficult to flush out parser bugs; to my knowledge, WLA's macro language
is not context-free.

That is more than probable. :)

>
> I suspect parsing rules being applied incorrectly is the biggest source
of bugs. Did you ever write a grammar for WLA?

Haha no. :) WLA DX is the result of pure organic growth and two partial
rewrites. When I started making WLA DX I was learning C and I'm sure I had
no idea about writing grammars LOL. :) For years and years I was the only
one coding, didn't get many reports etc. mail about it, and here we are. :)

nicklausw:
Wait, what do you mean by a "grammar"? Like, a proper parser?

vhelin:
I think something like this:

  https://en.wikipedia.org/wiki/Formal_grammar

Yeah, well, rules for the parser. It's been possible for years, I think, to
first create the grammar for a parser, and use a specific code generator to
create (C? C++?) code that parses such data/"language". I didn't use
anything like that when making WLA DX, just incrementally added new rules
and stuff to the parser / assembler...

Man the hot weather here in Philippines (Thailand was equally hot) is still
draining my batteries - evenings just fly by and I really can't get
[...]

PR #556, opened 2023-02-04
s_inz is now l_inz

jeffythedragonslayer:
I got some warnings that `l_inz` wasn't initialized when I made it local, so I just initialized it to zero.

### DIFF
--- a/decode.c
+++ b/decode.c
@@ -41,8 +41,6 @@ extern int g_use_wdc_standard;
 extern int g_input_number_expects_dot;
 #endif
 
-static int s_inz;
-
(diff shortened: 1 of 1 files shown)


vhelin:
I think it's better that I handle s_inz as sometimes it's value needs indeed to be static in the file, sometimes it can be replaced with a local variable. This is one of the great leftover things from my youth when I was still learning C, I deeply regret reusing the same variable...

vhelin:
Let's continue this conversation here: https://github.com/vhelin/wla-dx/issues/23

PR #559, opened 2023-02-21
full expanded name of assembler

jeffythedragonslayer:
"Wzonka-Lad" is one syllable shorter than "WLA" so I think the assembler will be easier for people to talk about if the full expanded name is put in a more obvious spot.


vhelin:
I'd rather have this in the form:

WLA DX (Wzonka-Lad Assembler Deluxe)
===========================

Remember to use as many ='s as there are letters

PR #566, opened 2023-02-24
pass -> phase

jeffythedragonslayer:
let's see what the tests say...


historical/README
```diff
@@ -3558,7 +3558,7 @@ WLALINK (the linker).
 
 WLA has four separate phases, called a little bit incorrectly passes:
```
vhelin, line 3559: After this PR they are called phases so they are not incorrectly called passes any more :)

vhelin:
I'm kinda ready to accept this PR as four phases are more correct than four passes. But don't be surprised if I one day decide to rename phase_1.c to pass.c... I'd like to sit on this at least a day.

historical/README
vhelin: Just this might be enough:

"WLA has four separate phases:"

:)

vhelin:
Ok I'm happy with this PR. If there are problems on e.g. the Amiga side I'll fix them...

PR #567, opened 2023-02-24
actually remove the makefile from VCS

jeffythedragonslayer:


vhelin:
Windows x64 and Windows x86 pipelines don't use cmake AFAIK, and ./run_tests.sh there builds byte_tester using that makefile that was removed in this PR.

Also Windows MSVC doesn't use cmake AFAIK, but it doesn't run ./run_tests.sh either...

PR #570, opened 2023-02-26
definition of compulsory directive

jeffythedragonslayer:
I didn't mention that the compulsory directives mentioned at the top must appear in that order in source code because I'm not sure if that also applies to `.ROMBANKS`, `.ROMBANKSIZE`, and `.ROMBANKMAP`, but it's a start.

### DIFF
--- a/doc/asmdiv.rst
+++ b/doc/asmdiv.rst
@@ -1,6 +1,24 @@
 Assembler Directives
 ====================
 
+A compulsory directive is one that must appear somewhere in 
+your source code, or else WLA-DX will produce an error.
+
+Some directives are always compulsory, such as:
+
+.MEMORYMAP
+.BANK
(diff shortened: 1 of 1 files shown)


doc/asmdiv.rst
```diff
@@ -1,6 +1,24 @@
...
+.ORG
+
+These directives are compulsory unless .ROMBANKMAP is defined:
+
+.ROMBANKS
+.ROMBANKSIZE
+
+Other directives are not compulsory.
+
+====================================================================
```
vhelin, line 20: Hmm what's the meaning of this line?

vhelin:
Thank you for this - I almost forgot it, and decided, based on this, to write a small quickstart which might be a little bit easier to understand.

PR #585, opened 2023-06-28
Improved makefile dependency rules generation to support deleted include files

mdagois:
Option -M generates makefile dependency rules. However, those rules break when an include file is deleted.

For example, assuming a file `main.asm` includes two other files `h1.inc` and `h2.inc`, the following makefile dependency rule would be generated.

```
main.asm: h1.inc \
  h2.inc
```

This works fine as long as the headers are there. However, if `h1.inc` is deleted (and the include statement removed from `main.asm`), `make` will fail, because `h1.inc` will still be referenced by the dependency rule, but there are no rule to produce `h1.inc`. The same issue would happen if `h2.inc` were to be deleted.

To solve the issue, the dependency rules are generated as follows.

```
main.asm: h1.inc
h1.inc:
main.asm: h2.inc
h2.inc:
```

Adding empty rules for each header prevents `make` from failing, and once compilation is over, any reference to the deleted include files will be gone.

### DIFF
--- a/include.c
+++ b/include.c
@@ -511,11 +511,10 @@ char *get_file_name(int id) {
 
 
 /* converts filename to forward slashes for make compatibility */
-void print_file_name(FILE *f, char *prefix, char *file_name) {
+void print_file_name(FILE *f, char *file_name) {
 
   char c;
   
-  fprintf(f, "%s", prefix);
   for (c = *file_name++; c != 0; c = *file_name++) {
     if (c == '\\')
       fputc('/', f);
@@ -524,14 +523,25 @@ void print_file_name(FILE *f, char *prefix, char *file_name) {
   }
 }
 
+void print_makefile_rule(FILE *f, char *target_file_name, char *prerequisite_file_name) {
 
-int print_file_names(void) {
+  print_file_name(f, target_file_name);
+  fputc(':', f);
+  fputc(' ', f);
+  print_file_name(f, prerequisite_file_name);
+  fputc('\n', f);
+
+  print_file_name(f, prerequisite_file_name);
+  fputc(':', f);
+  fputc('\n', f);
+}
+
+int print_file_names(char *target_file_name) {
 
   struct incbin_file_data *ifd;
   struct file_name_info *fni;
   struct stringmaptable *smt;
   struct string *fopens;
-  int is_first_line = YES;
   
   fni = g_file_name_info_first;
   ifd = g_incbin_file_data_first;
@@ -541,30 +551,25 @@ int print_file_names(void) {
   /* included files */
   /* handle the main file name differently */
   while (fni != NULL) {
-    if (is_first_line == YES) {
-      print_file_name(stdout, "", fni->name);
-      is_first_line = NO;
-    }
-    else
-      print_file_name(stdout, " \\\n\t", fni->name);
+    print_makefile_rule(stdout, target_file_name, fni->name);
     fni = fni->next;
   }
 
   /* incbin files */
   while (ifd != NULL) {
-    print_file_name(stdout, " \\\n\t", ifd->name);
+    print_makefile_rule(stdout, target_file_name, ifd->name);
     ifd = ifd->next;
   }
 
   /* stringmaptable files */
   while (smt != NULL) {
-    print_file_name(stdout, " \\\n\t", smt->filename);
+    print_makefile_rule(stdout, target_file_name, smt->filename);
     smt = smt->next;
   }
 
   /* filenames used in .fopens */
   while (fopens != NULL) {
-    print_file_name(stdout, " \\\n\t", fopens->string);
+    print_makefile_rule(stdout, target_file_name, fopens->string);
     fopens = fopens->next;
   }
 
--- a/include.c
+++ b/include.c
@@ -15,7 +15,7 @@
 #include "printf.h"
 
 
-extern int g_source_index, g_extra_definitions, g_parsed_int, g_use_incdir, g_makefile_rules;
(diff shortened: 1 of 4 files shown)


vhelin:
I have to admit that my makefile skills are a bit limited, so I don't 100% understand this. @maxim-zhao does this look good?

mdagois:
As an additional note, this is equivalent to gcc "-MP" option (https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html). I can add a new command line parameter if that's necessary.

maxim-zhao:
As I understand it, this is intended to make it so if you remove a file dependency from the WLA DX source, but don't re-run the makefile generation stage, then `make` will be satisfied. However, in my experience you always need to regenerate your makefile anyway, because it can just as much be missing newly-added dependency.

I don't think it will break my usage of this feature, so I think its OK to merge - although I also think it's a bad idea 😄

mdagois:
> As I understand it, this is intended to make it so if you remove a file dependency from the WLA DX source, but don't re-run the makefile generation stage, then `make` will be satisfied. However, in my experience you always need to regenerate your makefile anyway, because it can just as much be missing newly-added dependency.

No, there is a misunderstanding. The intent is to allow `make` to run the build rule, which outputs both the object file AND the dependency rules. You cannot run the build rule without regenerating the dependency file anyway if you use `-M`.

Right now, when you remove an `.include` and delete the corresponding file, the dependency rule generated from the previous build pass prevents `make` from executing the build rule again, because `make` does not know how to build the removed file (which is still in the dependency as this point). Adding a phony target basically tells `make` to not care about that missing file, which is fine, because it does not exist anymore. To fix the issue in the current system, you'd need to manually delete the dependency rule generated from the previous build pass (before the dependency was removed).
 
> I don't think it will break my usage of this feature, so I think its OK to merge - although I also think it's a bad idea 😄

It is not some shady change to support some weird use case. This option exists in gcc, clang and rgbasm. If you fear it could break existing build systems, I can definitely put it in its own option. Without that option though, the dependency rule generation in WLA loses a lot, if not all, of its appeal.

maxim-zhao:
I guess I'm confused then :) In my experience, you can't reliably combine makefile generation and object file generation, precisely because the latter depends on the former being correct. This option seems to allow one scenario (a dependency has been removed) to work, but then presumably the generated makefile will be updated and future builds won't have or need the phony targets.

I can see that this is a fairly widely used feature, although usually behind an extra option. However I don't see any issue with having it on by default, as it should make no difference to any other rules for the same files.

mdagois:
> I guess I'm confused then :) In my experience, you can't reliably combine makefile generation and object file generation, precisely because the latter depends on the former being correct. This option seems to allow one scenario (a dependency has been removed) to work, but then presumably the generated makefile will be updated and future builds won't have or need the phony targets.

Yes, with this rule, you can actually combine makefile rules and object generation in one single step.
Scenarios where the dependency rules are wrong correspond to the cases where the source file was touched anyway, so it should not matter that they are out of sync. And, as you mentioned, the dependency rules are updated to something correct after the build.
You do get that phony target that is useless most of the time, but it is hidden in a generated file, so it does not hurt.

> I can see that this is a fairly widely used feature, although usually behind an extra option. However I don't see any issue with having it on by default, as it should make no difference to any other rules for the same files.

I am not sure in what scenario the phony targets would get in the way. Most tools I know of have a separate option... Maybe I should add an option just in case?

mdagois:
Ok, just to be safe, I added an option (-MP) to enable the phony targets creation.
Also, I realized that gcc does not add a phony target for the main source file, so I change the code to do the same.
Please have a look.

vhelin:
Thanks!

maxim-zhao:
I have a project where I use makefile recipes with wildcards to cause dependencies to get automatically generated. The act of including a (non-existent) generated file in the assembly source causes the file to get generated based on the filename. If I were to try to build at the same time as makefile generation, the files would not get generated. Admittedly this is an edge case (although it was super useful to me). In the general case, I can see that folding them together might be helpful (although it currently then precludes any other text output at the assemble phase).

vhelin:
> I have a project where I use makefile recipes with wildcards to cause dependencies to get automatically generated. The act of including a (non-existent) generated file in the assembly source causes the file to get generated based on the filename. If I were to try to build at the same time as makefile generation, the files would not get generated. Admittedly this is an edge case (although it was super useful to me). In the general case, I can see that folding them together might be helpful (although it currently then precludes any other text output at the assemble phase).

I have to admit that I don't understand that if this is a a feature request or something else. :) I'm a totally beginner when it comes to makefiles...

mdagois:
> I have a project where I use makefile recipes with wildcards to cause dependencies to get automatically generated. The act of including a (non-existent) generated file in the assembly source causes the file to get generated based on the filename. If I were to try to build at the same time as makefile generation, the files would not get generated. Admittedly this is an edge case (although it was super useful to me). In the general case, I can see that folding them together might be helpful (although it currently then precludes any other text output at the assemble phase).

I don't fully understand the use case. However, you mention another issue I have with the current option: it output everything to stdout. The compilers I use the most gives you the option to output to a file.
It should be easy to add the option to allow that in WLA, as most fonctions take FILE pointers already. I will give it a try.

maxim-zhao:
It's OK - it's just an example of how makefile generation and complication/assembly might need to be separate stages. 

Makefile rules with wildcards can be used to implicitly run make "recipes" based on filenames seen in the build dependency tree. A common one is to make *.o be generated from *.c, for example. You can take this further by making the filename be "parsed" into args for generating the file, allowing you to simply mention the file at the point of usage, and have the recipe generate it. 

This is of course very  different to the usual pattern of creating a file and then using it. However it then means that the file is likely missing at the time of running the makefile generation - hence why WLA DX has special flags to ignore missing files when in makefile generation mode.

PR #588, opened 2023-06-30
Added option -MF

mdagois:
Added option -MF so that dependency rules can be output directly to a file. This also frees the stdout for other purposes. The option is only meaningful when using -M.


vhelin:
Thanks!

PR #601, opened 2023-08-17
Fix obsolete empty parens, caused error compiling on mac

thoolihan:
Fix obsolete empty parens, caused error compiling with homebrew toolchain

Error was "A function declaration without a prototype is deprecated in all versions of C"

More info here: https://stackoverflow.com/questions/75943541/a-function-declaration-without-a-prototype-is-deprecated-in-all-versions-of-c

### DIFF
--- a/wlalink/write.c
+++ b/wlalink/write.c
@@ -4603,7 +4603,7 @@ static int _labels_sort(const void *a, const void *b) {
 }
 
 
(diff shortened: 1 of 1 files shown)


vhelin:
Thanks! I'm aware of the issue, just thought that we didn't have "function_name()"s any more in the project... :)

PR #620, opened 2023-12-27
Fix for segfault in makefile generation mode

maxim-zhao:
Underlying cause was that the logic to make "fake" data for `.incbin` would reallocate a smaller buffer if a file was read later with a smaller required file size. The original change was over-eager to reallocate fake data buffers, and had another logic error.

### DIFF
--- a/include.c
+++ b/include.c
@@ -507,13 +507,17 @@ int incbin_file(char *name, int *id, int *swap, int *skip, int *read, struct mac
 
   if (g_makefile_rules == YES) {
     /* if in test mode, fake the data to be enough to read */
-    if (*read <= 0)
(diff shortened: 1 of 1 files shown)


vhelin:
I don't quite understand how this fix works, but I trust you know what you are doing. :)

maxim-zhao:
Basically, when in makefile mode, any attempt to open a file that doesn't exist will instead create and open a 0 byte temp file. My earlier change tried to then reallocate a buffer so that any skip/read args will succeed, but it mistakenly would do this even if that meant making the buffer smaller. The later code doesn't check sizes and then segfaults if the buffer is too small. This change makes it only enlarge buffers.

PR #636, opened 2024-09-15
Verbose mode when using -MF

mdagois:
Allowed verbose mode with -MF, as stdout is not used in that case to produce the makefile rules.


vhelin:
Thanks!

PR #637, opened 2024-09-17
Support for a symbol file format used by openMSX

mdagois:
The openMSX emulator debugger cannot read the NOCASH and WLA symbol file formats. Instead it uses a simple format using `equ` expressions.
For example, assuming a label `my_label` at address `$1234`, the corresponding entry in the symbol file would be `my_label: equ 1234H`.
In this PR, I added a new option to `wlalink`, `-sE`, to export a symbol file in the `equ` format.


vhelin:
Nice, thank you!

PR #639, opened 2024-10-20
Fixed the -MF option

mdagois:
When generating makefile rules, some file handling routines are skipped  or faked.
That's not the desirable behavior when using the -MF option, as it redirects the makefile rules generation to a different file, and the expection is that the build itself remains unchanged.
This PR fixes the -MF option behavior.

### DIFF
--- a/include.c
+++ b/include.c
@@ -16,7 +16,7 @@
 #include "main.h"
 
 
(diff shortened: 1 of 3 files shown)


vhelin:
Thanks!

Issue #658, opened 2025-05-11
.COMPUTESMDCHECKSUM is endian-dependent


michaelcmartin:
The Mega Drive checksum routine re-casts the ROM to an array of `unsigned short int` and then sums that. While this is the checksum algorithm, the algorithm is intended to be executed on a 68000 and should be treating the numbers as big-endian. When run on a little-endian architecture such as x86_64, the answer will generally be wrong.

The source below produces a minimal test ROM that covers both general operation and the case where one of the bytes has to carry (and thus causes the sums to stop being byte-swapped versions of themselves):

```
DEFAULTSLOT 0
SLOT 0 $0 $1000
SLOT 1 $FF0000 $1000
.ENDME
.ROMBANKSIZE $1000
.ROMBANKS 1

.SECTION "InitVectors" ORGA 0 FORCE
	.dd 0,RESET
.ENDS

.SMDHEADER
        SYSTEMTYPE "SEGA GENESIS    "
        ROMADDRESSRANGE $0, $0FFF
        EXTRAMEMORY "  ",$20,$20,$20202020,$20202020
.ENDSMD

.SECTION "Main" ORGA $0200 FORCE
RESET:	bra.b	RESET
	.dw	$a000
.ENDS

.COMPUTESMDCHECKSUM
```

vhelin:
Thank you for spotting and fixing the bug! I'll try to add this test into the tests folder later today...

PR #659, opened 2025-05-11
Fix endianness bug in COMPUTESMDCHECKSUM

michaelcmartin:
This ought to fix issue #658 by forcing big-endian math within the checksum at all times.


vhelin:
Thank you!

PR #722, opened 2026-06-12
Fix MC68000 labels and definitions named S0..S7 parsing as register A7

wbniv:
The new `SP` alias for address register A7 widened the operand parser's register condition to accept a leading `S`, but the digit branch was not gated to A/D. So a token like `S3` satisfied the `'0'..'7'` digit test and was then forced to A7 — meaning any label or definition named `S0`..`S7` used as an operand silently assembled as A7. For example:

```
move.w S3,d0   ->  30 0f      (i.e. move.w a7,d0)
```

instead of an absolute reference to `S3`.

Gate the digit branch with `c != 'S'` at all six MC68000 register-parse sites, so only the exact token `SP` maps to A7; `S3` now falls through to normal label/number parsing.

Adds `tests/68000/sp_alias_test`, asserting that a label `S3` and a definition `S6` resolve to their address/value while `SP` and `A7` still emit the A7 form.

### DIFF
--- a/decode.c
+++ b/decode.c
@@ -972,7 +972,7 @@ static int _mc68000_parse_ea(char *code, int *index, int *reg1, int *reg2, int *
       return FAILED;
     
     i++;
-    if ((code[i] >= '0' && code[i] <= '7') || (c == 'S' && toupper((int)code[i]) == 'P')) {
+    if ((c != 'S' && code[i] >= '0' && code[i] <= '7') || (c == 'S' && toupper((int)code[i]) == 'P')) {
       if (c == 'S')
         *reg1 = 7;
       else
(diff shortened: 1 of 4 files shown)


vhelin:
@wbniv The fix looks 100% solid, but can you remove that added test as it will fail as it uses S3 and S6 after the fix is applied? Then again the CI on this PR ran succesfully... ?

Too late for me to understand this, had a rough week. I'll get back tomorrow. :)

vhelin:
EDIT: S6 is a definition, but S3 should fail the test?

vhelin:
Argh, S3 is a label. :) Tired eyes...

Thanks!

PR #723, opened 2026-06-12
Fix mid() integer overflow that could crash the assembler

wbniv:
`mid()`'s out-of-range check computed `start + length > string_length`. For a large length — e.g. `mid(1, 2147483647, "abc")` — `start + length` overflowed to a negative value, the check passed, and the following copy loop wrote ~2 GB into the 256-byte stack item. A segfault driven purely by assembler source input.

Compare as `length > string_length - start` instead. At that point `0 <= start <= string_length` and `length >= 0` are already guaranteed, so `string_length - start` is a non-negative in-range value and the subtraction cannot overflow. `left()`/`right()` share the same check.

Adds `tests/6502/functions_error_test`, asserting the oversized `mid()` is rejected with a clean error. The test greps the log for the range message, which a crash would never print — so it genuinely fails against the unpatched (crashing) binary.

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -1475,7 +1475,7 @@ static int _parse_function_left_mid_right(char *in, struct stack_item *si, int *
     print_error(ERROR_NUM, "%s() start index %d is outside string \"%s\"!\n", name, start, tmp);
     return FAILED;
   }
-  if (start + length > string_length) {
+  if (length > string_length - start) {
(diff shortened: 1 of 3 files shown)


vhelin:
I'll check this out tomorrow, I apparently need sleep to continue. :)

vhelin:
Thanks, I'll merge this! But I need to remove the test as it will not work on e.g., Amiga OS.

Issue #649, opened 2025-01-15
[Error] .IF needs immediate data.


cyraid:
```
.MACRO BeginProc
.ENDM

.MACRO EndProc ARGS procName
  .IF procName > $FFFF
    rtl
  .ELSE
    rts
  .ENDIF
.ENDM
```

The assembler errors with ".IF needs immediate data" for the first line with the `.IF`. Why can't I use an `.IF` to compare with a label? Should just be a #? I've also tried with `\1` to no avail.

Test:
```
testFunc:
BeginProc
  lda #43
EndProc testFunc
```

In ASM16 I would use `$` to get the current PC Address to automatically detect a long jump if needed.

vhelin:
If you replace the call to EndProc with what you have in EndProc it looks to work if testFunc is outside .SECTIONs or inside a FORCE .SECTION, it seems. I guess calling a .MACRO with a label doesn't replace the label with its address, need to fix that...

Not sure I can fix this today, probably later this week...

cyraid:
Thanks for the quick response! Yes, that would be awesome. Thank you. :)

vhelin:
Please close this issue if that commit fixed it (I added a couple of tests that passed, but just to be sure...)

cyraid:
Hi @vhelin ! Thanks for the quick response. Unfortunately it doesn't seem to work for me. I'm using the 65816 SNES compiler. I've posted a sample GitHub Project here: https://github.com/cyraid/snesdevtest

The project is in vscode. To run do Terminal > Run Build Task.. In the tools folder it has the latest wla-65816 and wlalink.exe (x64 release, built with visual studio 2022), but you can feel free to use your own.

The `res/font.chr` is a chr file with fonts which is used to print on the screen via tiles and builds `out/test.sfc` when it's done.

`inc/common.inc` is where those functions are, and you can test them with the constant at the top `TEST_PROC_ADDRESS`, if it's 1 it will use the `.IF procName > $FFFF`, but if it's 0 it'll just use short jumps/returns.

`inc/snes_header.inc` is where I put the header specifics.
`inc/snes.inc` is the SNES specific constants
`/main.asm` is where the main code is.

If it passes all tests, when you run the sfc the result should print:
```
Init
Yes
Yes
Yes
Yes
Yes
```

I changed how the Procedures are defined, and I have to say, I like your Assembler as it allows this kind of flexibility:

```
;------------------------------------------------------------------------------;
[...]

PR #724, opened 2026-06-13
Fix section-relative labels passed to .MACRO resolving to section start

wbniv:
When a label defined inside a FORCE/OVERWRITE `.SECTION` is passed to a `.MACRO` and used in an assembly-time calculation, `_resolve_string` computed its value as the section's start address (`g_slots[...].address + section->address`) instead of the label's own address. Every label in the section therefore collapsed to the same value, so e.g. the difference of two such labels evaluated to 0.

**Introduced in** [`44ca5c12`](https://github.com/vhelin/wla-dx/commit/44ca5c127f8a16f060aaf2d3c14899c532bc969d) ("`.BASE`/`BASE` affects now labels when replaced with their address in `.MACRO` arguments", GitHub #649), which added both resolution branches at once — and only the in-section one used the wrong address:

```diff
+  /* a label outside .SECTIONs */
+  s->value = g_slots[dSI->slot].address + dSI->address;            // the label  (correct)
...
+  /* a label inside a FORCE/OVERWRITE .SECTION */
+  s->value = g_slots[section->slot].address + section->address;    // the section start  (bug)
```

The outside-section branch already used the label's own `dSI->address`; the FORCE/OVERWRITE branch used `section->address`. This fix makes the second match the first.

Use the label's address from the data-stream item (`dSI->address`, the absolute in-slot address) — mirroring the outside-section branch, which already uses `g_slots[dSI->slot].address + dSI->address`.

Adds `tests/65816/macro_section_label_test`, asserting that two labels at different offsets in a FORCE section, passed to a macro, yield their true distance rather than 0.

### DIFF
--- a/stack.c
+++ b/stack.c
@@ -4059,7 +4059,7 @@ static int _resolve_string(struct stack_item *s, int *cannot_resolve) {
               return FAILED;
             
             s->type = STACK_ITEM_TYPE_VALUE;
-            s->value = g_slots[section->slot].address + section->address;
+            s->value = g_slots[section->slot].address + dSI->address;
           }
           else {
             /* wla cannot resolve freely floating address labels -> only wlalink can do that */
--- a/tests/65816/macro_section_label_test/linkfile
+++ b/tests/65816/macro_section_label_test/linkfile
@@ -1,0 +1,2 @@
+[objects]
+main.o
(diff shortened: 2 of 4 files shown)


vhelin:
Looks good - thank you!

Issue #670, opened 2025-08-26
[Enhancement] Label substitution within .REPEAT


Ramsis-SNES:
Hi again,

I've been trying to use label substitution within a `.REPEAT` as outlined in chapter 4.8 of the documentation, where it states:

> It’s possible to substitute definition’s name with its value inside a
> label.
> 
> Here’s an example:
> 
>    .REPEAT 10 INDEX COUNT
>    Label_{COUNT}:                      ; -> Label_0, Label_1, Label_2...
>    .DW Label_{COUNT}
>    .ENDR
> 

Now, unfortunately I'm having trouble getting the labels I want (text pointers, in this case). I have multiple pointer blocks and, upon checking them, I realized it creates the same pointer values for all blocks, which leads me to believe that the `INDEX`, named `COUNT` in the above example, always starts at zero, even if I do, say, `.REDEFINE COUNT = 1024`  right before another pointer block.

It would be very, very handy if we could define our own starting value for the index, e.g.

```
.REPEAT 500 INDEX COUNT START 500
Label_{COUNT}:                      ; -> Label_500..Label_999
.ENDR
```
Or, if we wanted to get really fancy, even adding `STEP` like in `.ENUM`s: 

```
.REPEAT 500 INDEX COUNT START 1000 STEP 2
Label_{COUNT}:                      ; -> Label_1000, Label_1002 ... Label_1996, Label_1998
.ENDR
```
And, if that weren't enough already, maybe even allow `DESC` (or negative `.STEP`) as well ... :D

```
.REPEAT 100 INDEX COUNT DESC START 300 STEP 3
Label_{COUNT}:                      ; -> Label_297, Label_294 ... Label_3, Label_0
.ENDR
```
Thanks for considering this, but no worries if it's not feasible/too much work, you've already been doing so much for us all anyway :D 👍

Ramsis-SNES:
Works a treat, thank you so much, @vhelin :D 💯 👍

PR #725, opened 2026-06-13
Fix .REPT/.REPEAT hanging on an unexpected token after the count

wbniv:
The `.REPT`/`.REPEAT` `START`/`STEP` option loop only broke out on end-of-line: any other token after the count was read, rolled back, and retried on the next iteration — forever. So `.REPT 2 5` (or any stray token after the count) hung the assembler, and a parse failure printed the same error in an endless loop.

**Introduced in** [`b44d8341`](https://github.com/vhelin/wla-dx/commit/b44d8341a0cb33c5981108e6ae80a2714a9b5458) ("Added START and STEP to .REPEAT/.REPT", GitHub #670), which added the option-parsing loop. Its only exit is end-of-line; anything else is rolled back and the enclosing `while (1)` retries it indefinitely:

```diff
+      q = input_number();
+      if (q == INPUT_NUMBER_EOL) {
+        g_parsed_int = counter;
+        next_line();
+        break;
+      }
+      else {
+        /* this is not yet the end */
+        roll_back_to_remembered_source_file_position();   // → loops forever on a stray token
+      }
```

After the `INDEX`/`START`/`STEP` checks, require end-of-line: a parse failure returns `FAILED`, and anything else reports a clear "unexpected token" error and returns `FAILED`. Same-line `.REPT` bodies were already non-functional in this version (`.ENDR` continuation skips to the next line), so this only turns a non-terminating hang into a diagnostic.

Adds `tests/z80/rept_error_test`, asserting `.REPT 2 5` is rejected.

### DIFF
--- a/phase_1.c
+++ b/phase_1.c
@@ -10919,18 +10919,18 @@ static int _directive_rept_repeat_while(int is_while) {
         continue;
       }
 
-      /* test for EOL */
-      _remember_current_source_file_position();
-
+      /* the only thing that's allowed here is the end of the line */
       q = input_number();
       if (q == INPUT_NUMBER_EOL) {
         g_parsed_int = counter;
         next_line();
         break;
       }
+      else if (q == FAILED)
+        return FAILED;
       else {
-        /* this is not yet the end */
-        _roll_back_to_remembered_source_file_position();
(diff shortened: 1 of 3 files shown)


vhelin:
Looks solid - thanks! (again I need to remove that test)

